PR #11, opened 2023-09-16
added prettier config and removed other lock files (except bun-lock)

theVedanta:


tipiirai:
Thanks! I like how small the prettier config is

PR #17, opened 2023-09-16
fix: ensure compileFile handles cases where `dest` path does not exist

joevtap:
Using `node:fs mkdir()` with `{ recursive: true }` to create the destination path if it is provided but does not exist. Otherwise `bun compile` script throws an error since the `test/dist` folder is untracked by git.


tipiirai:
Looks solid. Thanks!

PR #18, opened 2023-09-17
fix: pass src to render the file

ezradiniz:
I was trying to render a file using `renderFile` and realized it wasn't passing `src` to render.


tipiirai:
Thanks!

PR #52, opened 2023-09-23
Use named constants for node types

K-Sato1995:
Cool stuff! Thought it might be better to use named constants for  node types in `createApp` to enhance readability and maintainability while going through the codebase.


tipiirai:
I'm familiar with this pattern, but I think it doesn't give any additional value to the numeric constants accompanied by comments. I think the constants actually add a little cognitive load, and surely increase the size of the file. What do you think about this?

K-Sato1995:
Thanks for the response! Yup fair enough👍! 
Gonna close it! Looking forward to the other tools 😄

PR #54, opened 2023-09-24
Feat: Linting Configuration

kon-pas:
## The Reason

I think it's very essential to have a linter, especially in an open source project. It greatly helps to keep a consistent syntax style across the project and maintain the code base. Both Prettier and ESLint are very common and used in most projects that I have seen, including `vue`, `vite`, `react`, `next`, `svelte`, etc.

## Features

Added new package scripts:
- `lint` - lints the whole source code and prints errors
```bash
bun lint
```

- `lint:fix` - formats the whole source code with changes printed via `lint`
```bash
bun lint:fix
```
(also considered the name `format`, but I found `lint:fix` to be more consistent with `lint`)

Also, having both Prettier and ESLint config files in the project mutes IDE errors, if one is using them (or any extension that depends on them)

> **NOTE**
> If it doesn't mute IDE errors, you may need to reload the IDE after pulling the changes.

I have also run the latter command and overlooked the syntax style changes. Beside the most basic spacing-related formatting, like empty lines, multiple spaces, tabs, etc, these are the significant ones that I have found out:

- Multiple methods chaining:
```js
// before
const ret = mods.map((key) => {
  const mod = MODIFIERS[key]
  if (!mod && key != 'once') keycode = key
  return mod
}).filter(el => !!el)

// after
const ret = mods
  .map(key => {
    const mod = MODIFIERS[key]
    if (!mod && key != 'once') keycode = key
    return mod
  })
  .filter(el => !!el)
```

- Multiple ternary operators chaining:
```js
// before
return key === $keys ? item :
  $keys.includes(key) ? item[key] :
  key == $index ? items.indexOf(item) :
  ctx[key]
  
// after
return key === $keys
  ? item
  : $keys.includes(key)
  ? item[key]
  : key == $index
  ? items.indexOf(item)
  : ctx[key]
```

- Method body wrapping:
```js
// before
shift() { arr.splice(0, 1) }

// after
shift() {
  arr.splice(0, 1)
}
```

- One line statements:
```js
// before
items = arrProxy(arr); repaint();

// after
items = arrProxy(arr)
repaint()
```

- Initialization of multiple variables at once:
```js
// before
let anchor, current, items

// after
let anchor,
  current,
  items
```

Also, these are the changes that are to be discussed - can be configurated, but I couldn't choose based on the project's style (used differently across the project):
- Trailing commas - I disabled trailing commas
- Line width - the standard is to use between 80 and 120, so I chose 100, the middle
- Quotes - most quotes are single `''`, but there are few double `""`, so I chose the former

## Changes

Installed and added the packages to dev dependencies:

- `prettier` - enables Prettier config and use
- `eslint` - enables ESLint config and use
- `eslint-plugin-prettier` - runs Prettier as an rule
- `eslint-config-prettier` - turns off all rules that are unnecessary or might conflict with Prettier
- `@babel/eslint-parser` - Babel's parser that supports all ECMAScript syntax (that ESLint doesn't)
- `@babel/core` - required by the above

Added Prettier and ESLint configs in the `.prettierrc.yaml` and `.eslintrc.yaml` respectively. Added scripts to `package.json` and also pushed the updated `bun.lockb`.

> **NOTE**
> All changes in files other than `.prettierrc.yaml`, `.eslintrc.yaml`, `package.json`, `bun.lockb` are from the linter.

I choose to use YAML because it's very minimalistic, though JSON is more common. Currently, I don't see the need to use `.cjs` or any other extension.

## Tests

The results of tests after making changes, including formatting:
```bash
15 pass
0 fail
89 expect() calls
```


tipiirai:
I also think it's important to keep a consistent syntax across the codebase! However, I'm hesitant to merge this one due to the following reasons

1. The project size goes from 4M to 40M (10x increase) and npm installs dev dependencies by default
2. The formatting options are very different from what I'm accustomed to

Is there a way to solve these?

tipiirai:
I think a better, and definitely more minimalistic approach is to establish good formatting rules and a matching `.prettierrc.csj` file. I started the work by updating this doc: https://github.com/nuejs/nuejs/blob/master/CONTRIBUTING.md

What are your thoughts?

kon-pas:
I didn't view dev dependencies as being that important to the project, because the final package that is shipped doesn't contain them, thus the size stays the same, but eventually I agree with you. The project is not that big, so we probably don't need them to be installed. It's all about the `lint` script, which we don't need that much.

Just having the `.prettierrc.csj` should be enough, but I would cover all [Prettier options](https://prettier.io/docs/en/options), because if we don't specify one, Prettier will use its default value anyway. I tried to match them all in `.prettierrc.yaml` of this PR, so you can just copy them and customize to your preferences.
File extension doesn't really matter. I thought YAML is the most minimalistic, but it's up to you. A full list can be found [here](https://prettier.io/docs/en/configuration).

> The formatting options are very different from what I'm accustomed to

Prettier is very opinionated. A formatting will always result with the code that I have enlisted before, because it can't be configurated (with Prettier).

Another solution that I would suggest is to just mute Prettier. This will prevent people from formatting the code with Prettier at all. We could have a `.prettierignore` with the whole project directory specified. [`/*`](https://github.com/prettier/prettier-vscode/issues/578#issuecomment-418767013) should do that.

tipiirai:
Nice! 

I definitely prefer YAML over the other options. 

Can we close this PR and do this step by step? This PR has so many things and I'm not comfortable with all of them. 

I thnk the first step logical is to just add the YAML file and make it match the current style of programming. Note: I can change any nonstandard habits!

kon-pas:
@tipiirai Sure!
I have just made a new PR #66 with only the Prettier configuration. Also, sorry for taking so long.

PR #66, opened 2023-10-01
Add: Prettier configuration

kon-pas:
Mapped all Prettier options to match the project's style, as discussed in PR #54.
Converted the configuration to YAML, for its simplicity.


tipiirai:
This is great! Thank you

PR #67, opened 2023-10-01
Fix Typo VARIABLE regular expression

YmBIgo:
Seems like typo in VARIABLE regular expression.

```javascript
[\$,a-z,_]
```

This codes allow `,` to be matched, and even if you want to include `,` you don't have to include `,` twice.

スクリーンショット 2023-10-01 15 00 41

----

For other codes, [here use 'map' for loop](https://github.com/nuejs/nuejs/blob/7945d8b5fd0b5850a0e56c1e70e8bc0cb2b79b5b/ssr/expr.js#L34) but you can use `forEach` instead, since forEach don't generate new array and therefore more cost effectively.


tipiirai:
Thanks! Nicely spotted!

PR #76, opened 2023-10-08
fix/resolve the bad Reg for {{{ size }}}  => { size }

Lonely-bear:
Before this Reg: 
"style="color: blue; font-size: {{{ size }}}px" => "{{{ size }" 
After:
"style="color: blue; font-size: {{{ size }}}px" => "{ size }"

### DIFF
--- a/ssr/expr.js
+++ b/ssr/expr.js
@@ -1,7 +1,7 @@
 
 const VARIABLE = /(^|[\-\+\*\/\!\s\(\[]+)([\$a-z_]\w*)\b/g
 const STRING = /('[^']+'|"[^"]+")/
-const EXPR = /\{([^}]+)\}/g
+const EXPR = /\{([^{}]+)\}/g
 
 
 // https://github.com/vuejs/core/blob/main/packages/shared/src/globalsWhitelist.ts


Lonely-bear:
I'm wrong, this RegExp don't effect the compiled result

Lonely-bear:
emm, It does affect, I have verified
image

tipiirai:
> It does affect, I have verified

@Lonely-bear so this is the expected behaviour and good to merge?

Thanks!

Lonely-bear:
> > It does affect, I have verified
> 
> @Lonely-bear so this is the expected behaviour and good to merge?
> 
> Thanks!

Yes, this regular expression affects the compilation result。It's good to merge。

Lonely-bear:
**expr.js**
````js
const EXPR = /\{([^}]+)\}/g
````
**islands.nue**
```html
<label :for="el, i in themes" :style="background-color: #{{{el.color}}};" class="theme">
     <input name="theme" type="radio" :checked="i === checkedIndex" @change="change(el, $event, i)">
     <h4>{ el.name }</h4>
</label>
```
Final **island.js**
image



**After Expr.js**
```js
const EXPR = /\{([^{}]+)\}/g
```
Final **island.js**
image

tipiirai:
Thank you!

PR #92, opened 2023-10-30
remove unused var

lorenzofox3:


tipiirai:
Good catch. Thank you!

PR #96, opened 2023-10-31
feat: RAF and time slicing

sullay:
Using RAF (requestAnimationFrame) and time slicing to perform component update operations.

1. Consider executing component updates in the RAF phase: 
By performing component updates within RAF callbacks, you can synchronize update operations with the browser's rendering, avoiding layout jank and improving the smoothness of animations.

2. Consider time-slicing based on device FPS: 
Since different devices have varying performance levels, adjusting time-slicing intervals based on device performance is essential to maintain a smooth application experience. This can be achieved through performance measurements and adaptive algorithms to ensure optimal performance across different devices.

It's possible that web-scheduler contains redundant logic that doesn't align with Nue.js' principle of simplicity. Providing a custom task scheduling feature may be a better choice.


tipiirai:
Hey. Thanks for the PR. Few questions / concerns here:

1. I was hoping to see discussion first before implementing major new features. Check [contribution guidelines](https://github.com/nuejs/nuejs/blob/master/CONTRIBUTING.md)
2. Would be important to see benchmarks/demo where this optimization new feature displays clear improvement
3. The web-scheduler dependency would make the client script many times larger. This is pretty severe. 
3. What does web-scheduler offer on top of the native `requestAnimationFrame`?

sullay:
Perhaps we should discuss in the "discussions" whether to introduce this new feature. 

Directly introducing web-scheduler is clearly not the most concise approach. This PR is only to verify the feasibility of introducing this feature with minimal effort.

Regarding explanation for question 3:

- web-scheduler will measure the device's refresh rate using the requestAnimationFrame API and use it as the basis for setting the time-slice length for task scheduling. 

- It manages tasks through a priority queue (implemented with skip lists and a hash map), running a certain length of time in each RAF cycle, with remaining tasks executing in the next RAF cycle.

tipiirai:
Hey. I think issues area makes more sense because the best features typically solve a problem. Here the (expected) problem is rendering performance — in which case there should be a demonstration of the bad performance. Feels like building a performance benchmarking tool is the first step. That is: we should focus on solving problems, not "verifying feasibility". 

What do you think?

tipiirai:
Btw: is there a good demonstration of web-scheduler somewhere? ideally something that shows how it works better than the native requestAnimationFrame.

sullay:
> Hey. I think issues area makes more sense because the best features typically solve a problem. Here the (expected) problem is rendering performance — in which case there should be a demonstration of the bad performance. Feels like building a performance benchmarking tool is the first step. That is: we should focus on solving problems, not "verifying feasibility".
> 
> What do you think?

I agree

sullay:
> Btw: is there a good demonstration of web-scheduler somewhere? ideally something that shows how it works better than the native requestAnimationFrame.

The main function of web-scheduler is to manage the execution timing of asynchronous tasks, breaking down a large task into smaller segments distributed across each frame to prevent a single frame from taking too long to execute. This can improve page responsiveness and reduce visual lag.

The animationFrameScheduler is based on requestAnimationFrame, and there is no straightforward comparison between the two.

As you mentioned, building a performance benchmarking tool should be the first step. Otherwise, it's challenging to assess the performance impact of the new feature on Nue.js.

PR #133, opened 2023-12-24
fix(nuekit-server): allow resolving index.html with base url

nobkd:
Fixes server not knowing index routes, e.g. `localhost:8080/blog` by adding `index.html` to links not having any extension, resulting in a file request to `localhost:8080/blog/index.html`


tipiirai:
Solid work again. Thanks!

PR #135, opened 2023-12-26
Windows adaptation: Fixed some error points in Windows systems

ghost:


tipiirai:
Oh gosh. I guess I merged these issues (after a lengthy Christmas break)  in the wrong order and this now causes a merge conflict. Can you check again? 🙏 This is super important because I cannot debug Windows issues personally!

fritx:
> because I cannot debug Windows issues personally!

I've tried to setup Windows CI, here is an WIP [preview](https://github.com/fritx/nue/commits/ci-win-only/) !

Before git cherry-pick this PR: https://github.com/fritx/nue/actions/runs/7399718947/job/20131776062
After git-cherrypick: https://github.com/fritx/nue/actions/runs/7399792114/job/20132011501

Seems the `YAML parse error in site.yaml` error gone with this fix.

However, the Windows CI test still got a lot to fix, seems like most of them are just path issues like:

image

See: https://github.com/fritx/nue/actions/runs/7399792114/job/20132011501  @tipiirai

packages/nuekit/src/init.js
```diff
@@ -22,7 +23,10 @@ export async function init({ dist, is_dev, esbuild }) {
     return await fs.stat(latest)
 
   } catch {
-    await fs.rmdir(outdir, { recursive: true })
+    try {
+      await fs.rm(outdir, { recursive: true })
+    } catch {}
```
fritx: Is this try/catch necessary? 
I cherry-pick this commit without the try/catch here, seems no problem here
See https://github.com/nuejs/nue/pull/135#issuecomment-1875652749

caesar1shi: When the target file does not exist, the program will crash on my end.

fritx: Makes sense. there are also lots of code like this: https://github.com/search?q=%22fs.rm%22+%22%7D+catch+%7B%7D%22&ref=opensearch&type=code

fritx: > When the target file does not exist

Also, should we pass a `force: true` to `fs.rm` ?
https://nodejs.org/api/fs.html
> force [`<boolean>`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures#Boolean_type) When true, exceptions will be ignored if path does not exist. Default: false.

packages/nuekit/src/site.js
```diff
@@ -84,7 +85,7 @@ export async function createSite(args) {
       return to
 
     } catch (e) {
-      if (e.errno != NOT_FOUND) throw e
+      if (e.errno != NOT_FOUND && e.errno!= ENOENT) throw e
```
nobkd:
```suggestion
      if (e.errno != NOT_FOUND && e.errno != ENOENT) throw e
```

packages/nuekit/src/site.js
```diff
@@ -99,7 +100,7 @@ export async function createSite(args) {
       !is_bulk && !self.is_empty && log(join(dir, base))
 
     } catch (e) {
-      if (e.errno != NOT_FOUND) throw e
+      if (e.errno != NOT_FOUND && e.errno!= ENOENT) throw e
```
nobkd:
```suggestion
      if (e.errno != NOT_FOUND && e.errno != ENOENT) throw e
```

packages/nuekit/src/site.js
```diff
@@ -127,7 +128,7 @@ export async function createSite(args) {
         })
 
       } catch (e) {
-        if (e.errno != NOT_FOUND) return console.error(e)
+        if (e.errno != NOT_FOUND && e.errno!= ENOENT) return console.error(e)
```
nobkd:
```suggestion
        if (e.errno != NOT_FOUND && e.errno != ENOENT) return console.error(e)
```

packages/nuekit/src/site.js
```diff
@@ -198,7 +199,7 @@ export async function createSite(args) {
         const html = await read(path)
         lib.push(...parseNue(html))
       } catch (e) {
-        if (e.errno != NOT_FOUND) {
+        if (e.errno != NOT_FOUND && e.errno!= ENOENT) {
```
nobkd:
```suggestion
        if (e.errno != NOT_FOUND && e.errno != ENOENT) {
```

nobkd (commented):
Looks great! I would only change some code style, as suggested here and maybe do fs.rm with force as noted by fritx

packages/nuekit/src/init.js
```diff
@@ -10,7 +10,8 @@ export async function init({ dist, is_dev, esbuild }) {
 
   // directories
   const cwd = process.cwd()
-  const srcdir = new URL('.', import.meta.url).pathname
+  const pathname = new URL('.', import.meta.url).pathname
+  const srcdir = pathname.startsWith('/') ? pathname.slice(1) : pathname
```
fritx: image

Seems this change can fail tests on non-Win OS

packages/nuekit/src/init.js
```diff
@@ -10,7 +10,8 @@ export async function init({ dist, is_dev, esbuild }) {
 
   // directories
   const cwd = process.cwd()
-  const srcdir = new URL('.', import.meta.url).pathname
+  const pathname = new URL('.', import.meta.url).pathname
+  const srcdir = process.platform === "win32" && pathname.startsWith('/')  ? pathname.slice(1) : pathname
```
nobkd:
```suggestion
  const srcdir = process.platform === "win32" && pathname.startsWith('/') ? pathname.slice(1) : pathname
```

tipiirai:
Thank you, guys!

PR #136, opened 2023-12-29
Fix cli init mkdir missing recursive: true and error message not showing

fritx:
Apperance:
`$ nue` does nothing but exits

Steps to reproduce:
```sh
cd create-nue/simple-blog
nue  # here nue is linked from my local nue directory (nuekit@0.1.6 - 310d51f)
```

Before fix:
image

After fix:
image


tipiirai:
Super. Thanks

PR #137, opened 2023-12-29
Fix back button popstate not working on the first pushState

fritx:
Apperance:
The back button does not work right after the first link navigation, but begins to work after the second navigation.
( Tried Chrome 120.0.6099.129 and Safari 16.6 (17615.3.12.11.3, 17615) )

Reason:
The first popstate e.state=null, and getting path=undefined
image

Steps to reproduce:
```sh
cd simple-blog
# both
nue
# and
nue build --production
nue serve --production
```

Fix Approach:
~~Record the `initial_path` and use it when path is undefined.~~

~~Or maybe, just `loadPage(location.pathname)`, regardless of e.state.path ? ref: https://github.com/fritx/silent/blob/2x/blog/vendor/blog.js#L409-L413~~

ref: window.onpopstate, event.state == null?
https://stackoverflow.com/questions/11092736/window-onpopstate-event-state-null

### DIFF
--- a/packages/nuekit/src/browser/page-router.js
+++ b/packages/nuekit/src/browser/page-router.js
@@ -69,6 +69,11 @@ export function setSelected(path, className='selected') {
 }
 
 
+// Fix: window.onpopstate, event.state == null?
+// https://stackoverflow.com/questions/11092736/window-onpopstate-event-state-null
+is_browser && history.pushState({ path: location.pathname }, 0)
+
+
 // autoroute / document clicks
 is_browser && onclick(document, (path) => {
   loadPage(path)


tipiirai:
Thank you @fritx. These fixes are great!

tanwwg:
I don't think it's fixed yet, I get null null whenever I hit the back button on 0.3.3.

fritx:
Still: window.onpopstate, event.state == null?
https://stackoverflow.com/questions/11092736/window-onpopstate-event-state-null

Looks like calling `history.pushState(initialState)` **twice** is more reliable. Thoughts? cc @nobkd 
I'm not sure as I tried it only once and it worked fine on my side.

PR #139, opened 2023-12-30
Fix all tests with `bun test`

fritx:
Before fix:
image

After fix:
image

Steps to reproduce:
```sh
cd nuejs/nue
bun install
bun test
```

Description:
1. Replace `..` with `../index.js` in `compile.test` and `render.test`, ref: https://github.com/nuejs/nue/commit/4e77356cbc487be8af2a55081e3a001b7dfef02f#r135937470
1. `write()` is async and should be awaited
1. [bun] `workspace` should be `packages/*` (consistent with pnpm-workspace.yaml), otherwise `bun/npm install` will not install dependencies of nuekit
1. [bun] `esbuild` is required while `bun test` and should be added to `nuekit/devDependencies`
1. [bun] bun.lockb changed (we can git-diff via some configs, ref: https://bun.sh/docs/install/lockfile)


nobkd:
https://github.com/nuejs/nue/blob/854f20f53aecd8f32d87860b760cd24b8f1a3209/packages/nuejs/test/compile.js#L2

This line above also uses `..`-import. It's probably better to be consistent and use `../index.js` everywhere. (Even if nothing fails with the current)
What do you think?

fritx:
> This line above also uses `..`-import. It's probably better to be consistent and use `../index.js` everywhere. (Even if nothing fails with the current) What do you think?

Excellent! I'm agreed. Just pushed again. @nobkd

tipiirai:
Super cool. Thank you

tipiirai:
Thanks. I am so happy to get these important fixes/additions.

tipiirai:
@fritx I was thinking of not introducing esbuild as a devDependency, because it's such a big package. Maybe have the test 
 task load esbuild asynchronously and if found, only then run the test. What do you think?

fritx:
@tipiirai my pleasure. Yes, I was considering it hard too.

I ended up with requiring esbuild in devDeps and the testing, because I found it already in pnpm-lock.yaml: 
https://github.com/nuejs/nue/blob/master/pnpm-lock.yaml#L31-L34

> Maybe have the test
task load esbuild asynchronously and if found, only then run the test.

IMO, in principle, we should test all the features that the library supports. But we should still try a better way to remove it from `devDeps` (to keep simplicity).

Maybe move it to `optionalDeps` and ~~find a way to manually turn it on in CI testing~~ ( [it's already there](https://github.com/nuejs/nue/blob/master/packages/nuekit/test/kit.test.js#L236) ) ?

PR #140, opened 2023-12-31
fix(page-router): don't prevent opening `_blank` target

nobkd:
Preventing `_blank` target results in no new tab and therefore no action.


tipiirai:
Cool. Makes sense. Thanks!!

PR #144, opened 2024-01-01
feat: title_template config key

nobkd:
This allows to make titles in the format you like, e.g. containing site name.

The config could look like this:

```yaml
# site.yaml
title_template: '%s | My Site Name'
```


tipiirai:
Nice. This was actually pretty clever. Thank you!

PR #147, opened 2024-01-02
fix: add type to favicon field

### LINKED ISSUE #146: NueKit doesn't Generate Type for Favicon
When you use the `favicon` property in `site.yaml`, if you use a `PNG` file, NueKit doesn't generate `type="image/png"` in the `<link />`, and so the favicon doesn't work.

nobkd:
resolve #146

Adds the image type to the favicon field

### DIFF
--- a/packages/nuekit/src/layout.js
+++ b/packages/nuekit/src/layout.js
@@ -1,4 +1,5 @@
-
+import { extname } from 'node:path'
+import { TYPES } from './nueserver.js'
 
 export function renderHead(data, is_prod) {
   const {
@@ -53,7 +54,7 @@ export function renderHead(data, is_prod) {
   if (components) pushMeta('nue:components', components.map(uri => `${base}${uri}`).join(' '))
 
   // misc
-  if (favicon) head.push(`<link rel="shortcut icon" src="${favicon}">`)
+  if (favicon) head.push(`<link rel="shortcut icon" type="${TYPES[extname(favicon).slice(1)]}" src="${favicon}">`)
 
   // inline style
   inline_css.forEach(css => head.push(`<style href="${base}${css.path}">${ css.content }</style>`))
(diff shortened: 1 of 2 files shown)


packages/nuekit/src/layout.js
```diff
@@ -53,7 +54,7 @@ export function renderHead(data, is_prod) {
   if (components) pushMeta('nue:components', components.map(uri => `${base}${uri}`).join(' '))
 
   // misc
-  if (favicon) head.push(`<link rel="shortcut icon" src="${favicon}">`)
+  if (favicon) head.push(`<link rel="shortcut icon" type="${TYPES[extname(favicon).slice(1)]}" src="${favicon}">`)
```
fritx, line 57: maybe should to lower case here? otherwise, .PNG doesn't work

tipiirai: Nicely spotted TYPES in nueserver.js! I think it's good to merge like this for now. I think we should add a getMime() method to nueserver later.

tipiirai:
Thank you!

packages/nuekit/src/layout.js
nobkd: > maybe should to lower case here? otherwise, .PNG doesn't work

Agreed!

> Nicely spotted TYPES in nueserver.js!

:wink: 

> getMime() method to nueserver

Sounds good

Not sure, but it would be probably good to expand the list with [more common types](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types). 
It was quite interesting looking at a PDF in text format on the browser. Even though it is just the dev server.

nobkd: > `getMime()` method to nueserver

Maybe even a package, which has all the up-to-date types like [`mrmime`](https://github.com/lukeed/mrmime) or [`mime`](https://github.com/broofa/mime)?

PR #150, opened 2024-01-03
Revert unnecessary changes of bun.lockb in PR #149

fritx:
Sorry for the unnecessary changes (bun.lockb) in last PR #149 , here is the **revert** >_<

Below is the `git diff bun.lockb` in this PR:

```diff
diff --git a/bun.lockb b/bun.lockb
index 38f7e48..853b438 100755
--- a/bun.lockb
+++ b/bun.lockb
@@ -1,6 +1,6 @@
 # THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
 # yarn lockfile v1
-# bun ./bun.lockb --hash: DEEABC817505CDA6-f41a17c91aabae47-E5034553CF1C77BA-d5d19ed64226cf44
+# bun ./bun.lockb --hash: 01C33369770C3C80-7bbaedb9804acfe5-FC545C77E671B54B-b6468c0bcba01140
 
 
 argparse@^2.0.1:
@@ -70,8 +70,24 @@ marked@^9.1.6:
   resolved "https://registry.npmjs.org/marked/-/marked-9.1.6.tgz"
   integrity sha512-jcByLnIFkd5gSXZmjNvS1TlmRhCXZjIzHYlaGkPlLIekG55JDR2Z4va9tZwCiP+/RDERiNhMOFu01xd6O5ct1Q==
 
-"nuejs-core@nuejs-core":
+nuejs-core@^0.3.0:
+  version "0.3.0"
+  resolved "https://registry.npmjs.org/nuejs-core/-/nuejs-core-0.3.0.tgz"
+  integrity sha512-8RTa3Hz+tE9dfeBe01+oT6kKhknMmETTPc3Ap1tURCmzNdvXamUFbt2FNaHi1OBmGWk0i8FYKFvhj6ZtCSpO9Q==
+  dependencies:
+    htmlparser2 "^9.0.0"
+
+"nuejs-core@packages/nuejs":
   version "workspace:packages/nuejs"
   resolved "workspace:packages/nuejs"
   dependencies:
     htmlparser2 "^9.0.0"
+
+"nuekit@packages/nuekit":
+  version "workspace:packages/nuekit"
+  resolved "workspace:packages/nuekit"
+  dependencies:
+    diff-dom "^5.0.6"
+    js-yaml "^4.1.0"
+    marked "^9.1.6"
+    nuejs-core "^0.3.0"
```

PR #154, opened 2024-01-09
Fix Windows CI with jest as alternative to bun, also add more matrix envs

fritx:
Goal:
Add and fix CI for Windows

Problem:
`bun` is not availble in Windows currently (ref: https://github.com/oven-sh/bun/issues/8045, https://github.com/oven-sh/bun/pull/7991).
Tried with both `setup-bun` ([docs](https://github.com/oven-sh/setup-bun?tab=readme-ov-file#usage), [failed](https://github.com/fritx/nue/actions/runs/7397873002/job/20125810699)) and `powershell -c "irm bun.sh/install.ps1|iex"` ([docs](https://bun.sh/docs/installation#windows), [failed](https://github.com/fritx/nue/actions/runs/7462503061/job/20304947574))

Solution:
For Windows, switching to `npm install` + `jest` as alternative to `bun install` + `bun test`.

Details:
- Completing the CI matrix.envs: `ubuntu-bun`, `ubuntu-jest`, `macos-bun`, `macos-jest`, `windows-jest`
- Making tests compatible with node + jest:
  - replacing `toInclude` => `toContain`
  - adding some missing `is_browser` detections
  - skipping `esbuild: false` test (bun only)
  - introducing [`es-main`](https://github.com/tschaub/es-main) to detect and prevent unexpected call of `getArgs` (throwing on the wrong args)
- Making tests compatible with Windows:
  - introducing [`toMatchOSPath`](https://stackoverflow.com/questions/67325342/how-to-run-os-agnostic-jest-test-files-that-check-paths)
  - handling some paths with `path.normalize` or `(path/posix).join`

✅ The CI are all passing:
https://github.com/nuejs/nue/actions/runs/7463368446

image


.github/workflows/test.yaml
nobkd, line 1: Just an idea... should we use the lock-file(s) for dependency installation?
That *might* make the test-runs more reproducible.
Referring to https://github.com/nuejs/nue/pull/150#issuecomment-1875224333


I mean... the test / build library might still change (new line 25 / 33), but at least the main dependencies would be the same

I'm not sure if it's a great idea, but it came to my mind.

fritx: Yes. It's a good question. I'm +1. For each project, we have different choices. Like:
- simple: totally ignoring all the lock-files
- sample: ...or just keeping one of the lock-files as a sample, e.g., yarn.lock (via bun-install -y)
- strict: ...or even keeping all the lock-files (bun, yarn, pnpm, npm, etc)

IMO, I personally prefer the choice 1 or 2, as nuejs is a quite lightweight framework or tool. Keeping more than 1 lock-files can be much effort while committing and code-reviewing. (which means to remove most of the lock-files and add them into .gitignore)

How do you guys think? @nobkd

fritx:
Sorry. Closing for some cleaning work. Would re-open later...

PR #155, opened 2024-01-10
Fix Windows CI via node+jest (as alternative to bun)

fritx:
#154 can not be re-opened, so created a new PR with less diff change:

Goal:
Add and fix CI for Windows

Problem:
`bun` is not availble in Windows currently (ref: https://github.com/oven-sh/bun/issues/8045, https://github.com/oven-sh/bun/pull/7991).
Tried with both `setup-bun` ([docs](https://github.com/oven-sh/setup-bun?tab=readme-ov-file#usage), [failed](https://github.com/fritx/nue/actions/runs/7397873002/job/20125810699)) and `powershell -c "irm bun.sh/install.ps1|iex"` ([docs](https://bun.sh/docs/installation#windows), [failed](https://github.com/fritx/nue/actions/runs/7462503061/job/20304947574))

Solution:
For Windows, switching to `npm install` + `jest` as alternative to `bun install` + `bun test`.

Details:
- Completing the CI envs: (ubuntu, macos, windows) x (bun, node+jest)
- Making tests compatible with node+jest:
  - adding some missing `is_browser` detections
  - introducing [`es-main`](https://github.com/tschaub/es-main) to detect and prevent unexpected call of `getArgs` (throwing on the wrong args)
  - using `jest-extended` to eliminate the `expect.*` matchers gap between `bun test` and `jest`
  - separating kit-init.test.js to run sequential, making tests passed (seems chdir is affecting the other tests)
- Making tests compatible with Windows:
  - introducing [`toMatchPath`](https://stackoverflow.com/questions/67325342/how-to-run-os-agnostic-jest-test-files-that-check-paths)
  - handling some paths with `path.normalize` or `getPosixPath`

PR #162, opened 2024-01-13
test(nuekit): drop removed function from tests

nobkd:
The tests fail, because a removed function `parseMarkdown` (now handled by `nuemark`) is imported by the test file `nuekit/test/misc.test.js`.

See the workflow run test log https://github.com/nuejs/nue/actions/runs/7500154128/job/20418283484#step:4:106

Should also solve the red badge in the README mentioned here: https://github.com/nuejs/nue/pull/155#issuecomment-1888722726

### DIFF
--- a/packages/nuekit/test/misc.test.js
+++ b/packages/nuekit/test/misc.test.js
@@ -1,7 +1,7 @@
 
 
 import { buildCSS, findModule } from '../src/builder.js'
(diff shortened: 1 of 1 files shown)


tipiirai:
Awesome. Thanks! I can see all tests passed for this commit. For some reason, the red badge is still there.

nobkd:
For me it's green again. Maybe it just took a while to update?

PR #164, opened 2024-01-13
chore: drop `Permalink for` for now

nobkd:
This is just a minimal typo fix.

But I see a problem with this approach, as the `title` property is hard-coded this way.

What if I want to create a Website not in English? (I do.) Then the `Permalink for 'x'` approach doesn't work anymore.

In conclusion, my concern is the localizability.

---

*Tests will run properly again after #162 is merged.*

### DIFF
--- a/packages/nuemark/src/render.js
+++ b/packages/nuemark/src/render.js
@@ -125,7 +125,7 @@ const renderer = {
     const rich = parseHeading(html)
 
     delete plain.text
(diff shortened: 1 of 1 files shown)


tipiirai:
Indeed. You are right. Perhaps drop the "Permalink for" part altogether at this point and fix this later. I'm thinking of a bigger project where all major localization-related things are implemented at once.

tipiirai:
Thank you!

Issue #168, opened 2024-01-17
Support for syntax highlighted code blocks


goblinfactory:
please provide native support for embedding code examples - ie. make nuejs a first class blogging platform for coders

I'd like to be able to support embedding code examples in markdown similar to how we do it in github.

For example, current nuejs has a bug, where collections are not sorted by date, so I tried to write an experimental blog post, **using nuejs** describing how I hacked the sorting as follows **_(surrounding the code block with triple ticks)_**

```html
<article @name=blog-posts>
  <script>
    constructor({ posts }) {
      this.items = posts.sort((a, b) => {
        const dateA = new Date(a.date.replace(/\./g, '/'));
        const dateB = new Date(b.date.replace(/\./g, '/'));
        return dateB.getTime() - dateA.getTime();
      });
    }
  </script>

<a :for="post,i in items" href={base}{post.url} class="flex post">
```
I can't even do it using `<pre>` or `<code>` ,  that can't work!

This currently renders as nothing [blank] in current version of nue. I think it ignores the triple tick and tries to compile it? Not sure. Either way, my Expectation; is for it to be rendered (htmlEncode'd) plus some default out of the box syntax colourising and formatting and developer appopriate out of the box styling, that is not affected by any other css site wide settings. 

Just my 2 cents. Currently I'm having to use screenshots to capture some thoughts and feedback on nuejs. Finding it difficult to actually blog about nuejs. So this might help make nuejs easier to write about.

goblinfactory:
This would also empower nuejs to become a great platform for document automation, or perhaps simply supplementation. Where the devs simply use it for writing documentation, because it's incredibly quick to write docs with it, AND more importantly, it keeps stying out of the way of content, so content can be easily edited, while remaining "in the flow", so ..perfect for coders. Let's replace confluence with nuemark, ...and keep documentation easily under source control, so that it's always in sync with what it's describing. That and later I'll bang on about C4 diagramming support, lol! #eatAnElephantOneBiteAtaTime

nobkd:
There is a possibility to highlight code with a function: https://nuejs.org/docs/reference/nuemark-api.html#highlighter
But I think it's not really accessible from Nuekit currently...

This info box states: https://nuejs.org/docs/concepts/nuemark.html#fenced-code-blocks
> A Nue-optimized syntax highlighter will be released on the next version of Nuemark.

tipiirai:
[...]

PR #169, opened 2024-01-17
fix(code-block): escape html chars to text content

nobkd:
Partial fix for #168

Escapes chars from code blocks to prevent them from being parsed as HTML.

This is not ideal, as code markup parsers have to then check for replaced char codes.

But replacing after running the `fn` function is also not great, because it potentially appends real HTML parts (e.g., `<span>`s) to do, for example, highlighting.

### DIFF
--- a/packages/nuemark/src/render.js
+++ b/packages/nuemark/src/render.js
@@ -59,10 +59,20 @@ export function renderLines(lines, opts={}) {
   return renderPage(page, opts)
 }
 
+const specialChars = {
+  '&': '&amp;',
+  '>': '&gt;',
+  '<': '&lt;',
+  '"': '&quot;',
+  "'": '&apos;',
+}
+
+const replaceSpecialChars = new RegExp(`[${Object.keys(specialChars).join('')}]`, 'g')
+
 export function renderCodeBlock({ name, code, attr }, fn) {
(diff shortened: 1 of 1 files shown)


packages/nuemark/src/render.js
```diff
@@ -59,10 +59,20 @@ export function renderLines(lines, opts={}) {
...
+  "'": '&apos;',
+}
+
+const replaceSpecialChars = new RegExp(`[${Object.keys(specialChars).join('')}]`, 'g')
+
 export function renderCodeBlock({ name, code, attr }, fn) {
   // console.info(name, code, attr, fn)
   if (name) attr.class = concat(`syntax-${name}`, attr.class)
-  const body = join(code)
+  const body = join(code).replace(replaceSpecialChars, c => specialChars[c])
```
fritx, line 75: should be `replaceAll`, here? otherwise, would only replace the first one.

tipiirai: Looks like there is a `'g'` flag on the RegExp which makes it global (replace all)

tipiirai:
Looks solid. I might later apply this regexp only if there is no custom syntax highlighter provided on the `fn` argument. I'm introducing a Nue-specific highlighter later.

PR #171, opened 2024-01-18
Fix parseExpr in case of single quote in str, and nueserver...

fritx:
1. Fix parseExpr in case of single quote in str
image

2. Fix ERR_HTTP_HEADERS_SENT if files like favicon.ico not found in nueserver
image


tipiirai:
Thanks!

PR #172, opened 2024-01-18
test(code-block): add test for html text escaping

nobkd:
Follow up to #169

I forgot to add a test, so...
Just a simple test to ensure, that escaping is performed properly for code blocks.

---

Just a question: Shouldn't fenced code blocks be structured like this?

```html
<pre>
  <code>
  // My code
  </code>
</pre>
```

Because currently, it just consists of a `<pre>`-tag. And I also only know it the way from my example.


tipiirai:
Thanks!

PR #185, opened 2024-02-01
Add pubDate to createFront and expand content collection test

goblinfactory:
- Added pubDate to createFront so that the `content-collection` test can actually assert on a collection with different dates to prove that the collection is being returned "newest" first based on RFC 822 `pubDate`  value.
- Test changed to use a manual inline snapshot so that the test can also serve as useful reference.

### DIFF
--- a/packages/nuekit/test/nuekit.test.js
+++ b/packages/nuekit/test/nuekit.test.js
@@ -45,8 +45,8 @@ async function getKit() {
   return await createKit({ root, dryrun: true })
 }
 
-function createFront(title) {
-  return ['---', `title: ${title}`, '---'].join('\n')
+function createFront(title, pubDate) {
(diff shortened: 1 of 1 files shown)


packages/nuekit/test/nuekit.test.js
```diff
@@ -45,8 +45,8 @@ async function getKit() {
   return await createKit({ root, dryrun: true })
 }
 
-function createFront(title) {
-  return ['---', `title: ${title}`, '---'].join('\n')
+function createFront(title, pubDate) {
+    return ['---', `title: ${title}`, `pubDate: ${pubDate ?? '2020-01-20'}`, '---'].join('\n')
```
nobkd, line 49:
```suggestion
  return ['---', `title: ${title}`, `pubDate: ${pubDate ?? '2020-01-20'}`, '---'].join('\n')
```

tipiirai:
Thank you!

PR #190, opened 2024-02-05
Fix failed test with createFront, fix #185

fritx:
Fixing #185

1. `First` should be passed through `createFront` also
2. `dir: 'blog/nested'` assertion failed in Windows, use OS-`join` instead

### DIFF
--- a/packages/nuekit/test/nuekit.test.js
+++ b/packages/nuekit/test/nuekit.test.js
@@ -121,7 +121,7 @@ test('get data', async () => {
 
 test('content collection', async () => {
   // Default sorting is on pubDate returning most recent first.
-  await write('blog/first-a.md', '# First')
+  await write('blog/first-a.md', createFront('First'))
   await write('blog/first-b.md', createFront('Second', '2020-01-04'))
   await write('blog/nested/hey1.md', createFront('Third', '2020-01-02'))
   await write('blog/nested/hey2.md', createFront('Fourth', '2020-01-03'))
@@ -133,8 +133,8 @@ test('content collection', async () => {
   expect(actual).toEqual([
     { url: '/blog/first-a.html', title: 'First', dir: 'blog', slug: 'first-a.html' },
     { url: '/blog/first-b.html', title: 'Second', dir: 'blog', slug: 'first-b.html' },
-    { url: '/blog/nested/hey2.html', title: 'Fourth', dir: 'blog/nested', slug: 'hey2.html' },
-    { url: '/blog/nested/hey1.html', title: 'Third', dir: 'blog/nested', slug: 'hey1.html' },
+    { url: '/blog/nested/hey2.html', title: 'Fourth', dir: join('blog', 'nested'), slug: 'hey2.html' },
+    { url: '/blog/nested/hey1.html', title: 'Third', dir: join('blog', 'nested'), slug: 'hey1.html' },
   ])
 })
 


packages/nuekit/test/nuekit.test.js
```diff
@@ -121,7 +121,7 @@ test('get data', async () => {
 
 test('content collection', async () => {
   // Default sorting is on pubDate returning most recent first.
-  await write('blog/first-a.md', '# First')
+  await write('blog/first-a.md', createFront('First'))
   await write('blog/first-b.md', createFront('Second', '2020-01-04'))
```
nobkd, line 125:
```suggestion
  await write('blog/first-b.md', createFront('Second', '2020-01-04'))
  await write('blog/first-c.md', '# Third')
```

Additionally, to the default fallback of that front-matter function (`2020-01-20`, see [line 49](https://github.com/nuejs/nue/pull/190/files#diff-c10aeaf246848a1abdde8aecf8d19aefb90b27781ed29d8f96e1957a1cb9dc96R49)), I would also add a test case where the `pubDate` field is `undefined`.

Also see https://github.com/nuejs/nue/commit/da78e4f71fb9c2d28ee7fb72765c6f98401ba37e#commitcomment-138287265

What do you think?

packages/nuekit/test/nuekit.test.js
```diff
@@ -133,8 +133,8 @@ test('content collection', async () => {
   expect(actual).toEqual([
     { url: '/blog/first-a.html', title: 'First', dir: 'blog', slug: 'first-a.html' },
```
nobkd, line 134:
```suggestion
    { url: '/blog/first-c.html', title: 'Third', dir: 'blog', slug: 'first-c.html' },
    { url: '/blog/first-a.html', title: 'First', dir: 'blog', slug: 'first-a.html' },
```

To fulfill `undefined` state at top of results.

packages/nuekit/test/nuekit.test.js
```diff
@@ -133,8 +133,8 @@ test('content collection', async () => {
...
     { url: '/blog/first-b.html', title: 'Second', dir: 'blog', slug: 'first-b.html' },
-    { url: '/blog/nested/hey2.html', title: 'Fourth', dir: 'blog/nested', slug: 'hey2.html' },
-    { url: '/blog/nested/hey1.html', title: 'Third', dir: 'blog/nested', slug: 'hey1.html' },
+    { url: '/blog/nested/hey2.html', title: 'Fourth', dir: join('blog', 'nested'), slug: 'hey2.html' },
+    { url: '/blog/nested/hey1.html', title: 'Third', dir: join('blog', 'nested'), slug: 'hey1.html' },
```
nobkd, line 137: Agreed :+1:

nobkd (changes requested):

packages/nuekit/test/nuekit.test.js
nobkd: Don't mind it, if you want to do that in #191

nobkd (approved):

packages/nuekit/test/nuekit.test.js
fritx: Agreed. Yes I think I've completed the test-cases in #191

fritx: One more reason I did not add such test-case to this PR was because that
Currently the order of arr returned by `site.getContentCollection` can vary and is not predictable when pubDate is not provided via `createFront`. See https://github.com/fritx/nue/actions/runs/7815178425/job/21317944828

If we want predictable results we have to fix/complete the sorting logic.
If we don't need predictable results we can not add such test-cases as they can fail in CI...
cc @nobkd

fritx:
UPDATE to https://github.com/nuejs/nue/pull/190#pullrequestreview-1863716574

One more reason I did not add such test-case to this PR was because that
Currently the order of arr returned by site.getContentCollection can vary and is not predictable when pubDate is not provided via createFront. See https://github.com/fritx/nue/actions/runs/7815178425/job/21317944828

If we want predictable results we have to fix/complete the sorting logic.
If we don't need predictable results we can not add such test-cases as they can fail in CI...
cc @nobkd

nobkd:
You're right. The sorting in your test is really inconsistent... Thanks for checking.

I thought, entries would just be put to the start of the list, when they miss the `pubDate`-key and order themselves by secondary sorting (e.g. title).

As that does not seem to be the case, I agree to (at least in this PR) not add tests, that contain this unpredictable results.

---

I think it would be nice, to have consistent and predictable sorting, but I'd recommend doing that in a separate PR (like #191 was planned to)

---

As this was supposed to just fix the tests, this should just get merged. :+1:

tipiirai:
Awesome. Thank you!

PR #191, opened 2024-02-05
Handle missing pubDate while sorting

fritx:
Commits based on PR #190 

1. Sort by title when pubDate is missing (pubDate defaults to -1 , comparing with existing dates). Thoughts?
2. Complete the test-cases for no front provided


fritx:
Converted to DRAFT because I get to realize that sorting by title is not good, the original order should be kept instead.

Issue #198, opened 2024-02-13
Allow users to pass additional IGNORE dirs and filenames from commandline 


goblinfactory:
Allow users to pass additional IGNORE dirs and filenames from commandline 

Likely affected code

https://github.com/nuejs/nue/blob/16b348a79a25c72ab3107ba0f50fd1ce56818d36/packages/nuekit/src/nuefs.js#L71-L79

**Rationale**

* I want to simplify the build steps for using Nuejs with Cloudflare by excluding the /functions directory.
* I can use Nuejs currently as-is but that would mean having 2 different folders for source code, one for website and one for functions. 

Yes, I know that sounds sensible, but that's not how Cloudflare works. They expect your "edge workers functions" to reside in a sub folder called /functions, meaning that I'd need to complicate the build step by building a folder, and the copying in the /functions directory post build. While I can do that, it's just more steps for a new person to Nuejs and new to cloudflare to trip over. 

- More importantly i suspect it would mean a file copy build step on every edit (tbc). Even if it doesn't it's just unecessary moving parts if the fix is a simple string value to ignore a folder.

- There are some other reasons as well, but this is the simplest explanation for now. I think this will be the first change that will require support user submitted args, and possible a change to `cli.js` _(code block below is truncated, you need to scroll to see the whole method.)_

https://github.com/nuejs/nue/blob/16b348a79a25c72ab3107ba0f50fd1ce56818d36/packages/nuekit/src/cli.js#L25-L62

**Tests**

New e2e test will likely go here, to test that when ignore values are provided, that the correct folders are ignored

https://github.com/nuejs/nue/blob/16b348a79a25c72ab3107ba0f50fd1ce56818d36/packages/nuekit/test/nuekit.test.js

goblinfactory:
**Update**
I've run a basic test on cloudflare and this doesnt appear to be interfering with cloudflare pages "functions".
I'd feel a lot more comfortable knowing the directory is phyiscally ignored. For now the work around would be to make sure you don't include any markdown files in the folders? 

The files inside the functions directory can't be served statically, since cloudflare will try to route requests matching file paths to matching code "handlers" and expect your code to handle them as REST requests. Which will just translate to GET requests that do or do not match any script files. There wont be any handlers matching the request, and I think what happens in that case, is the requests get converted into root requests for / and the paths get ignored.

For example, what I just tested;

I have a cloudflare function code at `functions/api/contact-us.js` that's hard coded to handl POST's and return json 
```json
{ 
[...]

PR #202, opened 2024-02-15
For discussion of issue #198 (dont merge!)

goblinfactory:
In order to add support for including optional IGNORE settings in `site.yaml`, It is necessary to make some rather large changes to 

- nuefs.js (specifically the critical `fswalk`)
- nuekit.js
-  site.js
- fswalk 

... due to the various hidden dependancies surrounging isLegit and the whole configuration merging.

In order to make the changes i've also had to move isLegit out to a separate file and create a factory method to return an "isLegit" function to avoid having to pass the "ignores" with every invocation.

I'm not comfortable that this is the right way to inject global configuration/settings, and that some of the existing bootstrapping might need discussing and re-thinking.

I've created  a PR so that these changes can be easily shared and discussed. 

### Changes to critical `fswalk`

This PR also shows that allowing user configurations in site.yaml related to files, ends up requiring delicate changes to `fswalk`, a critical piece of code that we need to limit changes to. (imho) 

### Hard code support for Cloudflare and leave user configuration for the next major release

Coverage is currently low; therefore for now, I recommend hard coding in support for excluding cloudflare worker functions folder, but leaving out the ability to customise it further with configurations in site.yaml.

I think a change to injecting global configuration settings requires a discussion, and might end up rethinking bootstrapping and be better to wait to include in a next major release? (hard coding solid good defaults for pit-of-sucess and idiot proofing the first version)

Hopefully the code changes in the discussion branch make sense; but it's a fairly chunky bit of changes, so I'll happy answer any questions if that will help.

I think that adding any other user overridable settings (in site.yaml) would end up requiring very similar changes, hence this PR for discussion only. 

Your thoughts?
Alan

p.s. I've got all the tests passing here. So while this branch is for discussion, it's passing all the tests.
I had to change the `beforeAll` and `AfterAll` to `beforeEach` and `AfterEach`, because some tests overwrite the site.yaml, and cause side effects. I though it would slow down the tests, but BUN is just unbelievably fast and didnt slow down at all. Having that nice clean separation for these integration tests is important.

After we've discussed this ... I'll create a new PR for hard coding support for ringfencing (protecting) cloudflare `functions` folder, and update the tests and bring in the `beforeEach` and `afterEach` change into that PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@
 node_modules
 dist
 .idea
+.vscode
+_test
--- a/packages/nuekit/src/isLegit.js
+++ b/packages/nuekit/src/isLegit.js
@@ -1,0 +1,15 @@
+/** Returns a function that will check if a file is not ignored and is legit.*/
+export function makeIsLegit(ignores = []) {
+    const IGNORE = ['node_modules', 'package.json', 'bun.lockb', 'pnpm-lock.yaml', ...ignores]  
+  
+    function ignore(name='') {
+      return '._'.includes(name[0]) || IGNORE.includes(name)
+    }
+    
+    function isLegit(file) {
+      return !ignore(file.name) && !ignore(file.dir)
+    }
+    
+    // TODO: real symdir detection
+    return isLegit
+  }
--- a/packages/nuekit/src/nuefs.js
+++ b/packages/nuekit/src/nuefs.js
@@ -1,6 +1,7 @@
 
 import { watch, promises as fs } from 'node:fs'
 import { join, parse, sep } from 'node:path'
+import { makeIsLegit } from './isLegit'
 
 /*
   Super minimalistic file system watcher.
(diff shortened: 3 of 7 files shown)


.gitignore
```diff
@@ -2,3 +2,5 @@
 node_modules
 dist
 .idea
+.vscode
+_test
```
goblinfactory, line 6: These 2 necessary to avoid checking in testing artifacts.

tipiirai:
Nice. I liked it! I was afraid of bloat, but it wasn't there! Not because of you, but it's just constantly everywhere and always.

goblinfactory:
@tipiirai So just to be clear, are you ok with the suggestion to NOT add "ignores" to site.yaml pre ver 1 or 2? 
Or does your comment mean you actually like this PR?
So that I'm not blocked by this, I'm going to do the hard coding anyway as a separate PR, and wait for your feedback on this PR. txs, A

goblinfactory:
closing; I still have this branch if we need to discuss. Issue #204 is a good work around.

Issue #220, opened 2024-02-24
Support for passing options to LightningCSS


nobkd:
I want to support all browsers, but use some modern CSS features like [`light-dark()`](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/light-dark#browser_compatibility), which is currently only stable in Firefox.

This is a rule, that can [get transpiled by lightningcss to be available in all browsers](https://lightningcss.dev/transpilation.html#light-dark()-color-function), if you [set a target](https://lightningcss.dev/transpilation.html#browser-targets). E.g.: `browser_targets: '>= 0.25%, last 2 versions, Firefox ESR, not dead'`

(I'm willing to implement that, as I already have a working draft, that I would just have to clean up. It would add an option to the config, whose name+structure would need discussion)

**Edit:** Relevant part:

https://github.com/nuejs/nue/blob/c1feaa81b00a1515f327b5a265a1f8ffd8cf0bdd/packages/nuekit/src/builder.js#L133-L135

```js
  try {
    const targets = browser_targets ? mod.browserslistToTargets((await findModule('browserslist', 'index.js')).default(browser_targets)) : null
    const include = mod.Features.Colors | mod.Features.Nesting
    return mod.transform({ code: Buffer.from(css), include, minify, targets }).code?.toString()
```

tipiirai:
Sounds good!

I have started thinking that Nue should support Lightning CSS natively, so it should be part of the dependencies. I think it's the best way to write CSS today. And having the ability to customize the supported features in the configuration would make sense, provided there are sane defaults. I think light-dark is a good default. 

Is there a way to make a solid configuration system without adding "browserslist" as a hard dependency?

nobkd:
I'm not quite sure, how to do that. Also, my approach wasn't feature based, but browser variant/query specific...

For example, if I have this `browserslist` query:

```
>= 0.25%, last 2 versions, Firefox ESR, not dead
```

it creates this array containing all versions and browsers, that should be supported:

```
[
  "and_chr 121", "and_ff 122", "and_qq 13.1", "and_uc 15.5", "android 121", "chrome 122", "chrome 121",
  "chrome 120", "chrome 119", "chrome 118", "chrome 116", "chrome 109", "edge 121", "edge 120",
  "edge 119", "edge 87", "firefox 123", "firefox 122", "firefox 121", "firefox 115", "ios_saf 17.3",
[...]

PR #230, opened 2024-03-14
feat: marked extensions

### LINKED ISSUE #182: Allow adding Marked Extensions in NueJS
> Hey I noticed nuemark uses marked as the underlying library. Is there an easy way to add marked extensions for things like [footnotes](https://github.com/bent10/marked-extensions/tree/main/packages/footnote)?

_Originally posted by @indifferentghost in https://github.com/nuejs/nue/discussions/159#discussioncomment-8126684_

I think this deserves its own issue for tracking.

nobkd:
- resolve #182 

**`WIP`/`POC`** implementation for marked extension loading.

## Description

The default file to load from is `marked.config.js` but this can be changed using the `marked_config` key in `site.yaml`.
The provided file path currently has to be relative to the `root` folder of the project / the root passed to `nue` command

The config `js` file has to be of format:

```js
// e.g. imports

export default [ // has to be an array
  // initialize marked extensions with functions / own objects here
]
```

If no file is found, the import() function silently returns an empty object which, on unpacking, results in an empty list for `marked_extensions` variable

## TODO

- [x] exclude marked config file from output folder
- [x] add tests

## Notes

- https://marked.js.org/using_pro#use

Maybe allow additional renderers with the following?

```js
marked.use({
 extensions: [
   //...
 ]
})
```


tipiirai:
Seems legit. This should only cover extensions that are not possible to implement as Nuemark extensions/tags. Footnote is probably the only one. Anything else?

nobkd:
Well, you can also add all the other marked extensions.
It is sometimes easier to just use an existing package instead of re-building stuff yourself.

e.g.
- extended tables syntax
- footnotes, smartypants (lite) (as it was removed from the base package)
- katex could also be interesting for some use cases
- etc.

https://marked.js.org/using_advanced#extensions

---

Still WIP, as I have to do the todos, and want to use `joinRootPath` (or changed name) to import the file.

---

BTW, I think, this config isn't needed, as the [options](https://marked.js.org/using_advanced#options) are deprecated (`smartypants`, `headerIds`, `mangle`) or nonexistent (`smartLists` at least it's not in this table):
https://github.com/nuejs/nue/pull/230/files#diff-9aaf075b4e5c776d59c27ef6c62e9c1079673f706482643126f348d7c6550cdfR9-R14

nobkd:
Rebased on #232 and removed deprecated options. Still have the todo's open...

Edit:
Notes:

- Seems like the path for `import` has to be absolute, as it probably tries to load relative to the current js file (just an assumption) / packages from node_modules folder
- I'm not really sure, how to append the marked config file to the list of files to be excluded from `dist`, because
  1. I don't quite know the location in code where this has to be added
  2. I have an absolute file path for importing, which could theoretically be absolute from the start

nobkd:
I've reached the conclusion, that for now, I'll just define the config file name to `marked.config.js`, as I am not too sure, how to add a dynamic relative or absolute path to the IGNORE field properly.
If you are against my current concept, please let me know.

(An alternative I thought of was to create a `nue.config.js` or similar, where multiple config keys are included, that might need some js import. But that's just a thought experiment)

The [current loading variant for marked extensions](https://github.com/nuejs/nue/pull/230/files#diff-9aaf075b4e5c776d59c27ef6c62e9c1079673f706482643126f348d7c6550cdfR38-R41) also does not allow changing the marked extensions at runtime, but I don't think that's a big problem.

---

> BTW, I think, this config isn't needed, as the [options](https://marked.js.org/using_advanced#options) are deprecated (`smartypants`, `headerIds`, `mangle`) or nonexistent (`smartLists` at least it's not in this table):

I have deleted that part from the code

---

I've also added `package-lock.json` and `yarn.lock` to the IGNORE field, even though it's unrelated to this PR. If you want to, I can also create a separate PR for this.

---

The only thing, that's left (for now), would be tests.

(Also haven't looked into the `Notes` part of my PR description, but the most relevant part is generally loading marked extensions)

nobkd:
Have to fix the test / some code...
The tests would fail, if #254 was merged already

nobkd:
Fixed the test, forgot some quotation marks... and for some reason the rendered md, is escaped, when run with nuekit.

Currently, the test is not run, because `process.exit()` which will be fixed with #254 
But I tested it locally with the line commented out.

This is now ready for review.

packages/nuekit/src/site.js
```diff
@@ -81,6 +81,9 @@ export async function createSite(args) {
     self.is_empty = true
   }
 
+  const markedConfig = joinRootPath(root, 'marked.config.js', true)
+  const { default: marked_extensions=[] } = await import(markedConfig).catch(() => ({}))
```
tipiirai: Oh. Didn't even know about this `default: ..` syntax. Nice one!

tipiirai:
Looks solid. I did [some changes](https://github.com/nuejs/nue/commit/c6fdd1b27539fb872a56e662cd8126b58b68b0ba) to how header id's are rendered on the dev branch. Can you check if this causes any conflicts when merging to dev? 

Thanks!

packages/nuekit/src/site.js
nobkd: Yeah. Renames 'default' to 'marked_extensions'  while unwrapping the variable from an object

nobkd:
I'm relatively sure, it will result in merge conflicts.
I moved the variable `renderer` and `marked.use` so the Nue markdown renderer is loaded before other extensions.
But it shouldn't be that much of a problem, since I only moved it inside the file without changing it.

nobkd:
Oooh. I think the test fails, because `nuemark` isn't the `link`ed variant from the current repo, but the latest version on the `npm` registry

If we run the following from the repo's root:

```sh
pushd packages/nuekit
bun link

pushd ../nuemark
bun link nuekit
bun link

popd
bun link nuemark

popd
```

Which links the both packages to each other and then `bun test`, then it should work...

nobkd:
^ the commit above (edit: now later in history) was needed, because I didn't run the other `inlineText` tokenizers anymore, when the `marked instance` was modified... (example is part of [marked-smartypants-lite](https://github.com/calculuschild/marked-smartypants-lite/blob/42a77d328b601a7cb272bddb88b53ed0b6ec5b9b/src/index.js))

---

@fritx do you know anything about this: https://github.com/nuejs/nue/pull/230#issuecomment-2042686220 ?
Shouldn't we always use the version on `GitHub` instead of the one from the `npm registry`, to have the up-to-date code tested?
Can you think of a simple way to link all the variants? Maybe adding the links to the root workspace is enough?

Edit: Also not quite sure, why the macos test was successful...

nobkd:
can we maybe fix this by replacing the `package.json`s in the sub packages `package-name@latest` with `workspace` stuff before installing? Will look around a bit.

EDIT 4:

version for both bun and node
```js
import { readdir, readFile, writeFile } from 'node:fs/promises'

const pkgs = (await readdir('packages', { withFileTypes: true }))
    .filter(dirent => dirent.isDirectory())
    .map(dir => `packages/${dir.name}/package.json`)

for (const pkg of pkgs) {
    const json = JSON.parse(await readFile(pkg, { encoding: 'utf8' }))
    for (const key of Object.keys(json.dependencies ?? {})) {
        if (key.startsWith('nue')) delete json.dependencies[key] // uses pkgs from top workspace node_modules
    }
    await writeFile(pkg, JSON.stringify(json,null,2))
}
```

Edit: btw does my extension loading not work with node for some reason. maybe I have to use a different file path variant....

Update: Added this idea to tests with #261
But we will have to check, if there isn't a better way. Because, this should also go for local development... (but i think the preinstall script of the base workspace was not enough, because the subpackages get initialized before the preinstall script?)
What I dislike with this approach is, that the files are changed and accidental commit to git can happen easily...

nobkd:
When I use the workspace packages (using changes from #261), the tests with `node+jest` (on `Linux`) still fail, somehow connected to the `dynamic import` with `absolute path` *not being found* in the test, **even though the file exists** (in package checks checking for the file fail too though (jest)...)

Bun in contrast resolves the absolute path correctly without problems. The same goes for npm if I link the nuekit package to some test project. I think, it just fails in the jest environment...

I haven't checked it with bun+jest yet. If it still occurs there, I can be sure that jest is the culprit. It might also be an `--experimental-vm-modules` issue...
Not sure, what to do here and how to check where this issue comes from (other than being related to dynamic `import()`)

Edit: I think `jest` does not like absolute paths and handles them as if they would have `/` as the project root instead of system root or so... but relative paths also don't work as I would expect...

nobkd:
I'm currently revisiting this, and slowly losing confidence of resolving this path-resolution issue or whatever this might be...

Any help or ideas are highly appreciated 😅

Thanks

nobkd:
It's such a dumb reason...

Jest seems to fail importing ESM `export`s from dynamic `import()`s in our setup
I should've just logged the errors caught on the dynamic import...

Trying to reproduce with a minimal example does not seem to work...

.... The project root has to be a module.... :facepalm:

nobkd:
I'm not sure, if the config file might be better to be implemented with name `nue.config.js` with option `marked` (or similar), so that other js-required options can be implemented, e.g. some `lightningcss` configs :shrug:.

nobkd:
Ugh, I can't pass the marked config to nuemark anymore, because the usage completely changed...
Will need to find a new way to pass it through.

Edit: Well, I just did a function that gets called to load the marked extensions. Is better anyways.

nobkd:
> [!note]
> I'm not sure, if the config file might be better to be implemented with name `nue.config.js` (or changeable in `site.yaml`) with option `marked` (or similar), so that other js-required options can be implemented, e.g. some `lightningcss` configs (see #220) :shrug:.

tipiirai:
I've been hesitant to merge this, even when completed, due to increasing dependency to Marked, which is not my favorite parser.

nobkd:
Closing because of upcoming `nuemark` version, that doesn't depend on `marked` anymore

PR #232, opened 2024-03-14
fix: use `cwd` + `nue root` as base for relative file output path

nobkd:
Allows to set paths relative to the root dir passed to the nue command instead of the cwd.
Also enables setting absolute paths

- [ ] Fix scripts not being copied/minified to the output folder :shrug:
- [ ] Tests, etc.
- [ ] There will always be something missing on some to-do list :)

Will have to investigate, why no `js` files are being copied.

EDIT: Ubuntu and Mac shouldn't be successful, as the paths are different from before...
Only Windows is correct with its test results...

### DIFF
--- a/packages/nuekit/src/site.js
+++ b/packages/nuekit/src/site.js
@@ -72,7 +72,7 @@ export async function createSite(args) {
     port = is_prod ? 8081 : 8080
   } = site_data
 
-  const dist = joinRootPath(root, rawDist || `.dist/${is_prod ? 'prod' : 'dev'}`)
+  const dist = joinRootPath(root, rawDist || join('.dist', is_prod ? 'prod' : 'dev'))
 
   // flag if .dist is empty
   try {
--- a/packages/nuekit/src/site.js
+++ b/packages/nuekit/src/site.js
@@ -1,5 +1,5 @@
 
-import { log, getParts, getAppDir, getDirs, colors, toPosix, sortCSS } from './util.js'
+import { log, getParts, getAppDir, getDirs, colors, toPosix, sortCSS, joinRootPath } from './util.js'
 import { join, extname, basename, sep, parse as parsePath } from 'node:path'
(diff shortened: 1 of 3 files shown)


tipiirai:
Are you trying to get rid of the [this chdir](https://github.com/nuejs/nue/blob/master/packages/nuekit/src/init.js#L34) command (which bothers me) or is this something else?

tipiirai:
Ahh... sorry. I did not concentrate enough. Looks like this relates to the `root` command line argument. Nice, makes sense. Is there a bug or the code is ugly?

nobkd:
The bug is, that you have to locate your output dir relative to the position of the place where the nue command is executed and not relative to the root, the nue command has been given

Edit: This is a bug, because it is unexpected, that the path is not located relative to the `site.yaml`, where you change the output path.

nobkd:
Still WIP. I need to do some local testing and maybe add a test case too.

Also, I'm not sure, if the tests should really pass currently...

nobkd:
Ok, fixed the tests. They shouldn't have passed on GitHub before... :shrug:
Maybe @fritx does have an idea why they did, even when they shouldn't have?

I think this is ready for review. I'm not *that* happy with the name of the helper function `joinRootPath`, so if you have a better name, let me know.

---

I think will rebase #230 on this, to simplify the logic for the following lines with `joinRootPath`: https://github.com/nuejs/nue/pull/230/files#diff-d08a333705c3aee86e4360625e0aa4e9a8be4ae5c86cbe642d0deecaf38c1868R83-R84

tipiirai:
Looks solid. Thank you!

fritx:
> Ok, fixed the tests. They shouldn't have passed on GitHub before... 🤷
> Maybe @fritx does have an idea why they did, even when they shouldn't have?

Looks like there are bugs in tests -- some tests are not running, due to a `process.exit()`.
I've figured it out while working on another PR...
Submitting a PR -- #254

PR #233, opened 2024-03-15
fix: reference component using attribute

### LINKED ISSUE #100: child component      name    Contains uppercase letters  ，error  not show   、ref  undefined
```
<!-- parent component -->

<div @name="nue_add">
    sum: { a } + { b } = <sum :a="a" :b="b"></sum>
    <br />
    Ab: { a } + { b } = <Ab :a="a" :b="b"></Ab>
    <script>
        a = 10
        b = 100
        mounted() {
            console.info(this.$refs['sum'])
            console.info(this.$refs['Ab'])  //show undefined
        }
    </script>
</div>

<span @name="sum">
    { a + b }
</span>

<span @name="Ab">
    { a } { b }
</span>
```

image

hichem-dahi:
Not sure if this is a fix or a feature.

Referencing components with custom tags using an attribute helps fix the issue of incorrect rendering when uppercase custom tags are present.

example (after parsing src):
`<Ab` --> `<ab ref-component="Ab">`

Fixes #100

### DIFF
--- a/packages/nuejs/src/browser/nue.js
+++ b/packages/nuejs/src/browser/nue.js
@@ -2,6 +2,7 @@
 import For from './for.js'
 import If from './if.js'
 
+
 const CONTROL_FLOW = { ':if': If, ':for': For } // :if must be first
 const CORE_ATTR = ['class', 'style', 'id']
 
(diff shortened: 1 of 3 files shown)


tipiirai:
This complicates the codebase quite a bit so I'd like to explore alternative approaches first. Personally, I never use uppercase letters in tag names. Maybe that could be a Nue best practice? What are your thoughts on that?

hichem-dahi:
It does seem like an expensive solution for a small issue. closing this for now.

PR #235, opened 2024-03-17
fix: accessing child's update function

### LINKED ISSUE #103: child   component  Does not contain script    $refs  update  is undefined
```

<div @name="nue_add" >
    { ty } 
    <br />
    <button type="button" @click="setty('a')">ty=='a'</button>
    <button type="button" @click="setty('b')">ty=='b' </button>
    <nue_child  :ty="ty"/>
    <script>
        ty = 'a'
        setty(v){ 
            this.ty=v;
            console.info(this.$refs["nue_child"])
            this.$refs["nue_child"].update();//Uncaught TypeError: this.$refs.nue_child.update is not a function

        }
    </script>
</div>

<div @name="nue_child">
    <span>{ty}</span>
    <span :if="ty==='a'">aaa</span>
    <span :else>bbb</span>
</div>
```

hichem-dahi:
Add `update` function to `impl` property even if the script tag is empty. 

Fixes #103

### DIFF
--- a/packages/nuejs/src/browser/nue.js
+++ b/packages/nuejs/src/browser/nue.js
@@ -249,15 +249,13 @@ export default function createApp(component, data={}, deps=[], $parent={}) {
       // setup refs
 
       // constructor
-      if (Impl) {
-        impl = self.impl = new Impl(ctx)
-
+      if (Impl) impl = self.impl = new Impl(ctx)
+        
         // for
         impl.mountChild = self.mountChild
         impl.$refs = self.$refs
         impl.update = update
-      }
-
+      
       walk(root)
 
       wrap.replaceWith(root)


hichem-dahi:
@tipiirai, can you help me add a test for this case?

tipiirai:
Looks solid. 

If anything, I'd like to have a solid server-side test runner for the client-side Nue JS library. Currently the test suite is client-side only and thus not ideal. Cannot be run with GitHub actions etc. I know, It's quite a big task to implement — but it's something that has to be done at some point when the single-page application pieces start to become relevant. 

Ideally the suite should be super lightweight based on a tiny JavaScript test library and not those monster Chromleess browser/server/cloud services.

PR #246, opened 2024-03-28
docs: update cli options help

### LINKED ISSUE #243: Update CLI help message to represent current options
https://github.com/nuejs/nue/blob/850e85fac1686567b33a3562f6058278d3043fc9/packages/nuekit/src/cli-help.js#L15-L18

Should contain all the stable options listed here:

https://github.com/nuejs/nue/blob/850e85fac1686567b33a3562f6058278d3043fc9/packages/nuekit/src/cli.js#L43-L55

E.g. `-I` can be really great, when you want to force refresh your `dist` output. Or `-s` if you run a production build and want to print the stats with the same command.

(If you want to, I can try making a PR with updated options + descriptions. I would leave `--push` out for now, as I think it was for not-ready-deploy-stuff?)

nobkd:
- resolve #243

### DIFF
--- a/packages/nuekit/src/cli-help.js
+++ b/packages/nuekit/src/cli-help.js
@@ -16,6 +16,10 @@ Options
   -r or --root          Source directory. Default "." (current working dir)
   -p or --production    Build production version / Show production stats
   -e or --environment   Read extra options to override defaults in site.yaml
+  -n or --dry-run       x
+  -s or --stats         Show site statistics after current command
+  -b or --esbuild       x
+  -I or --init          Force clear and initialize output directory
 
 File matches
   Only build files that match the rest of the arguments. For example:
--- a/packages/nuekit/src/cli-help.js
+++ b/packages/nuekit/src/cli-help.js
@@ -19,7 +19,7 @@ Options
   -s or --stats         Show site statistics after current command
   -I or --init          Force clear and initialize output directory
   -n or --dry-run       Show what would be built. Does not create outputs
-  -b or --esbuild       Use esbuild as bundler. Install esbuild manually
+  -b or --esbuild       Use esbuild as bundler. Please install it manually
 
 File matches
   Only build files that match the rest of the arguments. For example:
--- a/packages/nuekit/src/cli-help.js
+++ b/packages/nuekit/src/cli-help.js
@@ -16,10 +16,10 @@ Options
   -r or --root          Source directory. Default "." (current working dir)
   -p or --production    Build production version / Show production stats
   -e or --environment   Read extra options to override defaults in site.yaml
-  -n or --dry-run       x
   -s or --stats         Show site statistics after current command
-  -b or --esbuild       x
   -I or --init          Force clear and initialize output directory
+  -n or --dry-run       Show what would be built. Does not create outputs
+  -b or --esbuild       Use esbuild as bundler. Install esbuild manually
 
 File matches
   Only build files that match the rest of the arguments. For example:
--- a/packages/nuekit/src/cli.js
+++ b/packages/nuekit/src/cli.js
@@ -105,7 +105,7 @@ async function runCommand(args) {
   if (cmd == 'stats') await nue.stats(args)
 
   // build
-  else if (push || args.paths[0] || cmd == 'build') {
+  else if (dryrun || push || args.paths[0] || cmd == 'build') {
     const paths = await nue.build(args.paths, dryrun)
 
     // deploy (private repo ATM)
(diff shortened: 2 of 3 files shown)


tipiirai:
Thanks!

Is this still a work in progress? What are those "x" entries? If unaware:

--dry-run only shows what is being build, but does not do anything
--esbuild uses esbuild instead of Bun bundler for bundling/minifyin javascript

nobkd:
Yeah, it's work in progress. The `x`es are placeholders, because I haven't had read the code for those options. I also wanted to first check, if `esbuild` was a hard dependency now. If not, it would have to be mentioned in the description. Had no time yesterday and today, so the pr has to wait :)

nobkd:
What about the order of the options? Just did it now, as I thought would be relevant :shrug:

---

Also, I see some possible problems with the current option parsing:

https://github.com/nuejs/nue/blob/850e85fac1686567b33a3562f6058278d3043fc9/packages/nuekit/src/cli.js#L40

What if a file path for an `opt`ion or for `paths` starts with a `-`?
~~(Also paths with spaces, whether in quotes or not, won't work. If the `paths` are ever used by anyone)~~ (Edit: I think the shell should handle these)

---

https://github.com/nuejs/nue/blob/850e85fac1686567b33a3562f6058278d3043fc9/packages/nuekit/src/cli.js#L58

Probably won't ever happen, because the `if`-switch won't ever get there when another option was set.

`nue -r -s myroot somepath` -> would work, as root is taken, but is unexpected, because `root` comes after another option

`nue -e -s` -> won't error too, as another option was set, and the `opt` is never checked again

nobkd:
I don't think, it's correct that `-n` / `--dry-run` only works with `build` command set, and is ignored for no command, `serve` and `stats`

> [!important]
> Will have to add that to the description, if that's correct like that

Will have to check, if it's maybe just my current project...

nobkd:
Will have to check, if it's maybe just my current project...

nobkd:
My diff idea for this (https://github.com/nuejs/nue/pull/246#issuecomment-2027682311) would be

https://github.com/nuejs/nue/blob/850e85fac1686567b33a3562f6058278d3043fc9/packages/nuekit/src/cli.js#L39-L68

```diff
     // options
-    } else if (arg[0] == '-') {
+    } else if (!opt && arg[0] == '-') {

       // booleans
       if (['-p', '--production'].includes(arg)) args.is_prod = true
       else if (['-v', '--version'].includes(arg) && !args.cmd) args.version = true
       else if (['-n', '--dry-run'].includes(arg)) args.dryrun = true
       else if (['-h', '--help'].includes(arg)) args.help = true
       else if (['-v', '--verbose'].includes(arg)) args.verbose = true
       else if (['-s', '--stats'].includes(arg)) args.stats = true
       else if (['-b', '--esbuild'].includes(arg)) args.esbuild = true
       else if (['-P', '--push'].includes(arg)) args.push = true
       else if (['-I', '--init'].includes(arg)) args.init = true
 
       // string values
       else if (['-e', '--environment'].includes(arg)) opt = 'env'
       else if (['-r', '--root'].includes(arg)) opt = 'root'
 
       // bad options
-      else if (opt) throw `"${opt}" option is not set`
       else throw `Unknown option: "${arg}"`
 
     } else if (arg) {
       if (opt) { args[opt] = arg; opt = null }
       else args.paths.push(arg)
-    }
+    } else if (opt) throw `"${opt}" option is not set`
   })
 
   return args
 }
```

Problem is, that `option is not set` still does not happen, if we have a command not having any arguments left after an option requiring an argument e.g. cmd `nue -e` --> no args left, quits parsing

tipiirai:
Yeah. Clearly some issues in there with special file names and need to be fixed somehow. I think filenames starting with "-" are not a priority atm.

nobkd:
Okay... I started looking here again. And `--dry-run` only works with build / push / provided file to build:

https://github.com/nuejs/nue/blob/2842df8b1d20071e2bf60b905fc96df46b717a56/packages/nuekit/src/cli.js#L108-L118

https://github.com/nuejs/nue/blob/2842df8b1d20071e2bf60b905fc96df46b717a56/packages/nuekit/src/nuekit.js#L325

changing the latter to  `build([], args.dryrun)` would stop the files from being built at the start, but I think that is nonsensical.

I think it would be better to just add dry-run to `else if (dryrun || push || ...`. That behaves as being built, and prints what would be built without starting a dev server, that only builds after changing a file, and then only that changed file.

nobkd:
`--dry-run` is incompatible with `stats` as command and option.
If the `.dist` / output directory does not exist yet, it even throws an error.

I'll just skip stats printing if dry-run is active, alright?

nobkd:
I've changed the code according to my ideas. Please let me know if you see potential for improvement.

tipiirai:
Looks solid. Thank you!

PR #237, opened 2024-03-20
fix: fix an error when the project was started for the first time

nusr:
Steps to reproduce

1. run `bun create nue@latest` create a project
2. cd project directory
3. run `nue` start the project
4. report error message 

Screenshot 2024-03-20 at 22 36 23

PR #254, opened 2024-04-07
CI: Fix some tests are not running

fritx:
Commits:
- Fix missing `toMatchPath` for Windows testing (missing in #232)
- Fix some tests not running (mentioned in #232), removing process.exit (added in #237) instead of terminate kit.serve properly while testing


tipiirai:
Looks solid. Thanks!

PR #261, opened 2024-04-11
test: use workspace package versions

nobkd:
<details>
<summary>Previous code description</summary>

1. list package dirs
2. for each dir's package.json
   1. load json
   2. iterate over dependencies
      1. check if name starts with nue
      2. if true, remove from dependencies -> uses workspaces variant
   3. write updated json

```js
import { readdir, readFile, writeFile } from 'node:fs/promises'

const pkgs = (await readdir('packages', { withFileTypes: true }))
    .filter(dirent => dirent.isDirectory())
    .map(dir => `packages/${dir.name}/package.json`)

for (const pkg of pkgs) {
    const json = JSON.parse(await readFile(pkg, { encoding: 'utf8' }))
    for (const key of Object.keys(json.dependencies ?? {})) {
        if (key.startsWith('nue')) delete json.dependencies[key]
    }
    await writeFile(pkg, JSON.stringify(json))
}
```

---

Maybe I have it completely wrong, and the workspace versions are already used on tests, but I'm not too sure about that.

</details>

---

We will have to check, if there isn't a better way. Because, this behavior should also go for local development... (but i think the preinstall script of the base workspace was not enough, because the subpackages get initialized before the preinstall script?)

What I also dislike with this approach is, that the package files are changed and accidental commit to git can happen easily...

There's also the problem with the handling of releases.. There, proper versions should be used, but when some get updated.... We'll,  it sure. Lost my train of thought...


fritx:
Looks a little tricky, how do other people do for their monorepo projects?
Why doesn't "workspace" work in nuejs ? 😅

nobkd:
Thanks. I think, I just went straight down some path without looking around... 

I've checked some packages (e.g. nuxt, jest) and if they use some workspace package, they all use the `workspace:*` (or `workspace:^`  ?) syntax inside the nested `package.json`. But I've checked the same packages on npmjs.org and there they use numbered versions.
I don't know, if npm does that automatically, but  if not, this has to be changed before publishing.

I will check, if I find build / publish scripts that explain this. 

Edit: at least pnpm/yarn replaces them on publish. but using `workspace:*` would cause a fixed version (`X.Y.Z`). instead `workspace:^` should probably be used (`^X.Y.Z`) ([sadly isn't supported by bun yet](https://bun.sh/docs/install/workspaces))...

Edit2:

```
npm ERR! code EUNSUPPORTEDPROTOCOL
npm ERR! Unsupported URL Type "workspace:": workspace:*
```

looks like npm does not understand this workspace syntax. But I also can't raelly find any docs on npm's workspaces.... I might be easier to just use yarn/pnpm...

nobkd:
I'v now changed the installer to pnpm, because it understands the `workspace:*` syntax. Problem is, that pnpm uses the `X.X.X` version syntax on publishing instead of `^X.X.X` or `latest`.
But local development and tests should work properly now.

The code from before can be used to replace the packages back to `latest` if needed on local / for some publishing :shrug:.
Only change is 
```diff
-        if (key.startsWith('nue')) delete json.dependencies[key]
+        if (key.startsWith('nue')) json.dependencies[key] = 'latest'
```

fritx:
It's unlucky that npm doesn't support such syntax and standards if it is true.

I personally recommend CI to cover npm install, because npm is always the most basic compatibility object, although in many cases pnpm can bring significant optimization effects.

But if this update can solve the headaches you guys have been experiencing these days, I think we can just move forward. Cheers! 🍻

tipiirai:
Does this mean that `npm` no longer works after merging this PR?

nobkd:
<details>
<summary>Outdated</summary>

No. It would just not work for development of the nue packages. Users would still be able to use the published packages with npm, because the workspace version would be replaced on publish.

---

What still would be possible as alternative, is that we manually do all the linking commands.

I can also try to check in the npm code base, how its workspaces work, if there is any option like workspace:*...

The docs about npm's workspaces is really sparse, so that would be the only way I see, apart from manual linking.

I'd say, we try exploring both options, because I really don't like having to use pnpm / yarn instead of npm for dev apart from bun.

</details>

---

What I've read until now looks as if npm should resolve the [local workspace (sub-)packages before the ones on npm, as long as the semver version identifier is fulfilled](https://github.com/npm/cli/blob/9622597399ec93224fddf90a9209a98dbcfd6b2f/workspaces/arborist/docs/workspace.md). Maybe `latest` is a problematic one?

Edit: Yes it is (but I don't know why). For bot `bun` and `npm` the latest package is pulled from npm-registry first and not from local...

nobkd:
> [!important]
> `bun` and `npm` use the workspace package version when I use `*` instead of `latest`...
> But this is still not great for releases, because the latest cached version could theoretically be used of some package (with `*`) instead of the real latest one.
>
> Best would probably be real version specifiers like e.g. `^X.X.X`, but I think there was some reason that `latest` was used...
> @tipiirai do you remember why this was needed? Just for convenience? See https://github.com/nuejs/nue/commit/1c2e6e9951eba042eabcb3cdb6a20e103c313d49

tipiirai:
I started using `latest` after seeing issues from forgetting to update the latest version numbers from the package.json files. I guess "*" has the same effect than latest so this looks great to me.  

>  tried it again, and it didn't work again

So not safe to merge yet?

fritx:
I suddenly remembered how I handled such workspace case many years ago.
https://github.com/search?q=user%3Afritx+%22file%3Apackages%22&ref=opensearch&type=code

It works before, but I'm not sure if it will help with our current situation..

```js
// package.json
// "dependencies": {
"react-drag-sizing": "file:packages/react-drag-sizing",
"jsdom-innertext": "file:packages/jsdom-innertext",
"jsdom-selection": "file:packages/jsdom-selection",
```

UPDATE:

Here `npm install` linking is working again:

```sh
cd nuejs/nue
rm -rf node_modules
npm install
ls -l node_modules | grep nue
>> 
lrwxr-xr-x   1 fritx  staff   16  4 17 12:33 nue-glow -> ../packages/glow
lrwxr-xr-x   1 fritx  staff   17  4 17 12:33 nuejs-core -> ../packages/nuejs
lrwxr-xr-x   1 fritx  staff   18  4 17 12:33 nuekit -> ../packages/nuekit
lrwxr-xr-x   1 fritx  staff   19  4 17 12:33 nuemark -> ../packages/nuemark

ls -l packages/nuekit/node_modules | grep nue
>> 
lrwxr-xr-x  1 fritx  staff    14B  4 17 12:33 nue-glow -> ../../nue-glow
lrwxr-xr-x  1 fritx  staff    14B  4 17 12:33 nuemark -> ../../nue-mark
```

But I'm not sure if the dependency version is gonna be replaced properly when `npm publish`...

nobkd:
> tried it again, and it didn't work again

@tipiirai that was only, because I (tried `*` as version and) forgot to remove the `package-lock.json` after I checked if linking worked with `latest` as version (which wasn't the case).
So the current implementation does work properly and is ready to be merged, but I'm not sure, if it is that great, because cached versions could also be used with `*` instead of real `latest`... :shrug: 

@fritx The `file:packages/<dir>` sounds interesting and works with `npm`, but sadly does not with `bun`. And there would also be the issue with possible need to replace on publish.

In conclusion, I think, the `*` version is the best thing we can get without having to change the version for publishing or using a different package manager (I have tested `*` with npm, pnpm and bun).

Edit: I'm done now

tipiirai:
Makes sense. Thank you!

PR #273, opened 2024-05-02
chore: ignore coverage directory

nobkd:
When running the `jest` test command from GitHub actions, a `coverage` directory is created.
Command: `node --experimental-vm-modules node_modules/jest/bin/jest --runInBand --coverage`

This should not accidentally get pushed.


tipiirai:
Cool. Thanks

PR #268, opened 2024-04-27
chore(cli): make executable filtering more robust

nobkd:
Related:
- #239
- https://github.com/nuejs/nue/issues/253#issuecomment-2081230488

---

See `.bin` folder on Windows with `npm` and `bun` respectively
image  image

and on Linux with both `npm` and `bun`:
image

---

Close #253

Issue #253, opened 2024-04-06
It can not start the server with nuekit 0.5.3, but works with nue 0.4.3


xiejx618:
OS :   windows 11 23H2（22631.3374）
bun：1.1.2

E:\admin-dashboard>bun install nuekit@0.4.3 --global
bun add v1.1.2 (c8d072c2)

 installed nuekit@0.4.3 with binaries:
  - nue

 1 package installed [31.00ms]

E:\admin-dashboard>nue

✓ Nue 0.4.3 • Bun 1.1.2
✓ Initialize ./.dist/dev: ••••••••••✓ Serving site from ./.dist/dev
✓ Building site to: ./.dist/dev

✓ Single-page apps 1
   index.html

Time taken: 13ms

✓ http://localhost:8080/
^C
E:\admin-dashboard>bun uninstall nuekit --global
bun remove v1.1.2 (c8d072c2)

 - nuekit
 1 package removed [15.00ms]

E:\admin-dashboard>bun install nuekit --global
bun add v1.1.2 (c8d072c2)

 installed nuekit@0.5.3 with binaries:
  - nue

 1 package installed [3.32s]

E:\admin-dashboard>nue

✓ Nue 0.5.3 • Bun 1.1.2
✓ Initialize ./.dist/dev: ••••••••••
✓ Building site to: ./.dist/dev

Time taken: 1ms

E:\admin-dashboard>

kzt-wt:
I'm facing the same problem.

nobkd:
@kzt-wt are you also using `bun` on `windows`?

I'll test it on Windows later today / when I find the time. Then, we'll see, if there is some Windows related code / bun problems :)

kzt-wt:
Yes
I'm using bun on Windows.

nobkd:
I was able to reproduce this unexpected behavior. Will have to check since which change this happens. 
I also haven't checked yet, if it also happens for node on windows... (sorry for the slow progress, )

EDIT:
Happens for both node (with env shebang edited to node) and bun since nuekit 0.5.2
So 0.5.1 / 0.5.0 work fine, but some resources are not converted / found on serve I think...

nobkd:
I made a `git bisect` and the first bad commit is https://github.com/nuejs/nue/commit/6a7464ba33ce07a847ec954dc8979af7f04681e8 (Simplified stats command), even when fixing the doubled function `fmt` and the wrong imports.
Will investigate further.

nobkd:
[...]

PR #239, opened 2024-03-21
fix: fail to build static files on windows

nusr:
fix #179 

windows file separator is **\\\\**
POSIX  file separator is **/**
[https://nodejs.org/api/path.html#pathsep](https://nodejs.org/api/path.html#pathsep)

Issue #179, opened 2024-01-24
Nue didn't build static files on Windows using esbuild


brynne8:
The JS files are successfully built, while static html file not generated.

```
D:\git_work\simple-crm>nue build
✓ Initialize ./.dist/dev
   nuemark.js
   nue.js
   page-router.js
   app-router.js
   mount.js
   diffdom.js
   hotreload.js
   error.css
   favicon.ico

✓ Nue 0.3.2 • Node 18.19.0
✓ Building site to: ./.dist/dev

Time taken: 7ms
```

tipiirai:
How is the `nue` command using Node by default? Have you possibly edited the shebang?

brynne8:
> How is the `nue` command using Node by default? Have you possibly edited the shebang?

Yes. I edited the shebang. Since windows don't know `which`.

tipiirai:
Gotcha. Really hard for me to debug this issue without a Windows machine. Hope someone else can jump in.

PR #278, opened 2024-05-27
feat: support bundling of css `@import` rules

nobkd:
Support bundling of `@import` rules, useful, e.g. when importing CSS from `node_modules`

### TODO:

- [x] Write a simple test that involves `@import`

### DIFF
--- a/packages/nuekit/src/builder.js
+++ b/packages/nuekit/src/builder.js
@@ -1,8 +1,8 @@
 
 /* Builders for CSS, JS, and TS */
-
-import { Features, transform } from 'lightningcss'
-import { join, extname } from 'node:path'
+import { promises as fs } from 'node:fs'
+import { Features, bundleAsync } from 'lightningcss'
+import { join } from 'node:path'
 import { resolve } from 'import-meta-resolve'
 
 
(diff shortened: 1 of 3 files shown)


tipiirai:
Seems like a legit feature. One thing to note that hot-reloading is not supported for @import:ed files since Nue has no way of tracking those dependencies (without writing a Lightning CSS extension, but that feels like an overkill)

nobkd:
~~The test for Windows seems to fail, because of the code on master.
Here is the same run, but just for the branch (I reset to before the last merge commit, to be able to test without jest issues): https://github.com/nobkd/nue/actions/runs/9293144796~~

The thing about not being able to trace `@import`ed files for hot reloading should probably be mentioned in the docs then. This also probably only occurs, when the CSS file is in a local project, and not from a package... I agree, that it would be hard to track changes for imported files, and probably isn't needed.
We can still think about the possibility of being able to modify the lightningcss config that is passed to the transformation/bundling (#220)

nobkd:
@tipiirai (off-topic)

> [!important]
> It seems like you haven't got any pings for the `create-nue` repo for a long time.
> Please also look at the pull requests there, when you read this. :)
> 
> This nuejs/create-nue#43 blocks Windows usage.
> Same goes for `nuekit` (see #179 and #253 fixed with #239 and #268) because there was no release in a long time.

> [!Important]
> This should only get merged after #285 is merged somewhere (means, the tests are working!)

tipiirai:
@import sounds like the right solution to include Glow styling from node_modules, instead of hacking the dependency inside nuekit

nobkd:
> `@import` sounds like the right solution to include Glow styling from node_modules, instead of hacking the dependency inside nuekit

I'm not sure, how exactly you mean that :shrug:

But just, so you know: Imports from `node_modules` have to provide the path to/with `node_modules` (relative or absolute), it is not resolved automatically by lightningcss.

So I don't know how exactly you would do that when `glow` is just installed (globally) because of `nuekit` or locally, because `nuekit` is local, so...

tipiirai:
So the syntax of importing from node_modules is something like this ?

``` css
@import url("node_modules/nuekit/browser/syntax.css")
```

And the imported CSS is always bundled inside the parent CSS regardless of where it is imported?

The above would allow me to get rid of the automagical syntax.css auto-inclusion:

https://github.com/nuejs/nue/blob/master/packages/nuekit/src/site.js#L279

Thanks!

nobkd:
I'm not sure. Probably. It's too long in the past that I did the code changes. Also I not completely sure, what exactly you want to do. 

You probably can use the resolvePath from init.js to get an absolute path to the css.
But what's next? 

---

~~BTW: I forgot to change the merge request branch to point to master. So the changes from here are on dev now, but not on master.~~ (edit: merged dev to master)

PR #280, opened 2024-06-11
Render components inside of buttons

### LINKED ISSUE #277: Nuemark: How to add svg inside a button
### Describe the Bug

Using Nuemark, I can't get this example to work. The only way I found to get around it is to make the button a div, but that is not the behaviour I'm after.

This doesn't work:
[button]
    [mysvg]
    Learn More

This works but isn't a button:
[button]
    [mysvg]
    Learn More

### Environment

Example:
- OS: Arch Linux
- Package Manager: bun 1.1.8
- JS Runtime: bun 1.1.8
- Nuekit Version: nue 0.5.3

### Minimal Reproduction
1. Create a new skeleton project using `bun create nue`
2.  Add an svg of choice inside of layout.html, and give the svg a `@name="mysvg"`
3. Inside index.md, add a the code from the example that doesn't work

korabdoroci:
As the title says this simple change will allow you to put components from layout.html or another .nue file inside of buttons inside of your markdown. Fixes #277

### DIFF
--- a/packages/nuemark/src/tags.js
+++ b/packages/nuemark/src/tags.js
@@ -28,7 +28,7 @@ export const tags = {
   button(data, opts) {
     const { attr, href="#", content=[] } = data
     const label = parseInline(data.label || data._ || content[0] || '')
-    return elem('a', { ...attr, href, role: 'button' }, label || _)
+    return elem('a', { ...attr, href, role: 'button' }, content[0] ? nuemarkdown(content[0]) : label || _)
   },
 
   icon({ _, icon_base='/img', alt }) {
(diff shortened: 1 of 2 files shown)


nobkd:
@Korab-Doroci 
Can you create a test, that checks, whether this creates the expected results, please? Thanks.

korabdoroci:
As mentioned before, it will only work if the svg or other component is inside a .nue file. With this change it will not work if the component is inside the layout.html.

tipiirai:
I'm thinking of a `[link]` tag (or similar) for this use case. I think button should not allow any markup inside since the HTML spec only allows [phrasing content](https://developer.mozilla.org/en-US/docs/Web/HTML/Content_categories#phrasing_content)

nobkd:
Closing, as this should work in the upcoming nuemark version.
See following (dev) test case:

https://github.com/nuejs/nue/blob/e207424e6b89dc88b43442dbb294a366b0bd322f/packages/nuemark/test/tag.test.js#L220-L223

PR #281, opened 2024-06-15
chore: drop bun chdir hack

nobkd:
As far as I have seen, this works without the hack now.

This should also allow `bun` tests on Windows (See oven-sh/bun#11904 for chdir issue in tests)

Test run: https://github.com/nobkd/nue/actions/runs/9532359784

EDIT: I should try running this with a real code base, and not just with the tests
Edit 2: As far as I can see, this works just fine

### DIFF
--- a/packages/nuekit/src/cli-help.js
+++ b/packages/nuekit/src/cli-help.js
@@ -1,5 +1,4 @@
-
-import { colors } from './util.js'
+import { colors, openUrl, getVersion } from './util.js'
 
 const HELP = `
 Usage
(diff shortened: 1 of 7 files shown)


nobkd:
> [!Important]
> This should only get merged after `dev` is merged into `main` OR merge *this branch* into `dev` after #284 is merged (means, the tests are working!)

nobkd:
based on #309
theoretically can change related lines back, or this gets merged after #309

tipiirai:
Another awesome contribution. I'm glad you understand the codebase so well

PR #285, opened 2024-07-27
test: fix DOM tests, create-cli test

nobkd:
TODO:
- [x] rebase on dev
- [x] fixing DOM tests (wrapped main content with article tag)
- [x] find reason for `DOMException: This element has no parent node.` (was mounting components twice)

Other:
- simplify running tests
  - with Node: `npm test` or `bun run test`
  - with Bun: `bun test`
  - allow testing single package (when located at a sub-package's root)
- update checkout & bun action version


nobkd:
> [!Important]
> It's probably easiest to merge *this branch* into `dev` first (or into `master` after `dev` is on `master`), and then resolve the other tests, but feel free to choose your order.

---

If needed:
Changing target branch of a pull request:
`Edit title`, then there should be a branch selector

nobkd:
Will rebase on dev (since the tests there are fixed now) and then try to fix the DOM test.

Currently it looks like there is some mistake in `view-transitions.js`, as I've tested the example with the local nuekit code in a browser and the page content doesn't change to the new content until I force refresh. (same on navigating back after that.) I'll try at least locating the problem there.

I've also tested it in the browser with the nue(kit) version on npm, (with router option in site.yaml) and there the page changes, but the component on the page after navigation fails to mount.

I'll notify you, when I find something or get stuck.
I'll try to get started with it later today / tomorrow.

(I'll probably also try to change some code in the DOM test. E. g. use a http server instead of manually resolving the js, etc.) 

---

Update:

```js
11 |      *
12 |      * @param message Message.
13 |      * @param name Name.
14 |      */
15 |     constructor(message, name = null) {
16 |         super(message);
             ^
DOMException: This element has no parent node.
      at new DOMException (/path/to/nue/node_modules/happy-dom/lib/exception/DOMException.js:16:9)
      at replaceWith (/path/to/nue/node_modules/happy-dom/lib/nodes/child-node/ChildNodeUtility.js:27:19)
      at replaceWith (/path/to/nue/node_modules/happy-dom/lib/nodes/element/Element.js:446:26)
      at mount (/path/to/nue/packages/nuekit/test/page-router-test/.dist/dev/@nue/nue.js:340:12)
      at /path/to/nue/packages/nuekit/test/page-router-test/.dist/dev/@nue/mount.js:36:46
```

This error message is unrelated to the failing test. The `view_transitions` just does not work with the structure of `main` in `packages/nuekit/test/page-router-test/index.html` and `packages/nuekit/test/page-router-test/page/index.html`

When I add `$('body').innerHTML = $('body2', dom).innerHTML` to `loadPage()` in `packages/nuekit/src/browser/view-transitions.js` (e.g. line 20) the test works.
Alternatively, when I wrap the content of `main` with an `article`, it works too.

nobkd:
> [!important]
> Please read

As you can see, I wrapped the content of `main` with an `article` for now, as `view_transitions` seems to only support `article` or `aside` directly in `main` for now.
When this limitation gets lifted, this can be changed back.

---

Happydom doesn't seem to like, when the component gets mounted twice (reason, comment below)
(it then throws stuff, as in above hidden comment)

---

Components were mounted twice when there's `view_transititons` enabled:

See: https://github.com/nuejs/nue/blob/c92290632a498819d3efead07994b9b5bcc3f7ab/packages/nuekit/src/browser/mount.js#L59-L63
and: https://github.com/nuejs/nue/blob/c92290632a498819d3efead07994b9b5bcc3f7ab/packages/nuekit/src/browser/view-transitions.js#L205-L207

I dispatch the route event for DOMContentLoaded now in mount, which lets the route event handler mount the components.

---

The test error is, because something on dev changed.

nobkd:
now based on #312

nobkd:
> [!important]
> How to `tar gzip` a directory's content without dot-files:
> https://github.com/nuejs/nue/pull/309#issuecomment-2286220558

nobkd:
Is ready for review.

Includes also changes to create cli command.

Will probably have to check if -C automatically creates the dir
Edit: it doesn't. Will have to ensure dir exists beforehand.

tipiirai:
Awesome work again!

PR #287, opened 2024-08-05
Update css-best-practices.md

seneca:
First read of www2 docs.


nobkd (approved):

packages/nuejs.org/docs/css-best-practices.md
```diff
@@ -23,7 +23,7 @@ These are all described in this document. Once you harness the power of CSS you
 
 
 ## Respect constraints { #constraints }
-Limit yourself to as few fonts, font weights, colors, variables, elements, class names, and components as possible. This is beneficial for several reasons:
+Limit yourself to as few fonts, font weights, colors, variables, elements, class names and components as possible. This is beneficial for several reasons:
```
tipiirai: My current standard is to use the "oxford comma", or is that a bad practice?

packages/nuejs.org/docs/css-best-practices.md
```diff
@@ -118,7 +118,7 @@ article {
 }
 ```
 
-Simple selectors make easy-to-read and easy-to-maintain CSS, keep your file sizes to a minimum, and use lower specificity for overrides.
+Simple selectors are easy-to-read and easy-to-maintain. Keep your file sizes to a minimum and use lower specificity for overrides.
```
tipiirai, line 122: Your sentence reads better, but changes the meaning. ie. good CSS --> good selectors

packages/nuejs.org/docs/css-best-practices.md
seneca: This is a minefield or course but you usually put a comma before and when it’s connecting two independent clauses. It’s almost always optional to put a comma before and in a list. I would suggest to adhere to that rule.

packages/nuejs.org/docs/css-best-practices.md
seneca: Yes you are right. What about this new change:
Simple selectors make your CSS easier to read and easier to maintain. They keep your file size to a minimum and use lower specificity for overrides.
You can also use easy in the sentence. Your choice.

nobkd:
```suggestion
Simple selectors make your CSS easy to read and maintain. They keep your file size small and allow using lower specificity on overrides.
```
?

Edit: updated with proposal from next comments

seneca: another way of saying it would be:
They keep your file size to a minimum and allow you to use lower specificity for overrides.

nobkd: @tipiirai 

Do you agree with [this suggestion](https://github.com/nuejs/nue/pull/287#discussion_r1705705790) or would you like some changes?

You can just edit my comment if you want to do some changes and then commit the suggestion if you like it.

seneca: How about this below. If you don't like it then I am very happy with your suggestion.
They help keep your file size small and allow you to use lower specificity when making overrides.

packages/nuejs.org/docs/css-best-practices.md
tipiirai: Makes sense! I trust you on this one.

btw: I really appreciate you working on the grammar!

packages/nuejs.org/docs/css-best-practices.md
nobkd: Edited the code piece again.

PR #290, opened 2024-08-05
Update hello-world.md

seneca:
Minor changes

PR #289, opened 2024-08-05
Update global-design-system.md

seneca:
Please review these changes just in case.

PR #288, opened 2024-08-05
Update custom-layouts.md

seneca:
First read of www2 docs with proposed changes.

PR #292, opened 2024-08-05
docs: fix docs

nobkd:
These are mostly typo / grammar / link fixes, as well as markdown formatting.
I also added some TODOs for links, which are not contained in the docs (e.g. todomvc example)

I think this can be merged after 

- [x] #287
- [x] #288
- [x] #289
- [x] #290
- [x] #291
- [x] #293 
- [x] #294 
- [x] #295 
- [x] #296 
- [x] #297 
- [x] #298 
- [x] #299 
- [x] #300 
- [x] #301 
- [x] #302 
- [x] #303 

are done, as they are each smaller in scope, also because they're only one file each.

Some of my changes will probably already be fixed then, so I will always rebase, when any of them are merged.


packages/nuejs.org/docs/reactivity.md
```diff
@@ -34,11 +34,20 @@ view_transitions: true
...
+  opacity: 1;
+
+  &::backdrop {
+    background-color: #0005;
+    backdrop-filter: blur(4px);
+    transition: .5s;
+  }
 
   /* styles when the popover is about to be displayed */
   @starting-style {
```
nobkd, line 53: Changed, because *order matters*.

> [!Note]
> The `@starting-style` at-rule and the "original rule" have the same [specificity](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity). To ensure that starting styles get applied, include the `@starting-style` at-rule after the "original rule". If you specify the `@starting-style` at-rule before the "original rule", the original styles will override the starting styles.
>
> From: https://developer.mozilla.org/en-US/docs/Web/CSS/@starting-style

packages/nuejs.org/blog/tailwind-vs-semantic-css/index.md
```diff
@@ -93,7 +92,7 @@ The two metrics that measure page rendering speed are [first contentful paint](/
 
 
 
-Please compare [Tailwind metrics](//pagespeed.web.dev/analysis/https-spotlight-tailwindui-com/cqtnf4xxoy?form_factor=mobile) with [Semantic CSS metcis ](//pagespeed.web.dev/analysis/https-nuejs-org-spotlight/6nnhwwnz8b?form_factor=mobile).
+Please compare [Tailwind metrics](//pagespeed.web.dev/analysis/https-spotlight-tailwindui-com/cqtnf4xxoy?form_factor=mobile) with [Semantic CSS metrics](//pagespeed.web.dev/analysis/https-nuejs-org-spotlight/6nnhwwnz8b?form_factor=mobile).
```
nobkd, line 95: These links do not work anymore, and the page-speeds are negligibly different:
Nue: https://pagespeed.web.dev/analysis/https-nuejs-org-spotlight/i0itoo400b?form_factor=mobile
Tailwind: https://pagespeed.web.dev/analysis/https-spotlight-tailwindui-com/98p017y0jm?form_factor=mobile

Previous comment: https://github.com/nuejs/nue/commit/89aaa518494962c9bdf1f83e67f7c5a7d5b2eb4e#r144877042

---

The previous links were broken, probably, because the links are thrown away after some time.
The results can probably be saved with e.g. a local `lighthouse` installation (via bun / npm), and then the result file can get served.
AFAIK this generates a single html file, so it shouldn't be to hard.

packages/nuejs.org/docs/tags.md
```diff
@@ -219,9 +219,9 @@ Options given as YAML:
 
   `poster` a URL for an image to be shown before the playback starts
 
-  `preload` a [hint to the browser][preload] on what to load prior playback
+  `preload` <span>a [hint to the browser][preload] on what to load prior playback</span>
```
nobkd: Would break because of flex on `.options` otherwise, as a new element is inserted, and flex breaks the text wrapping (most noticeable on small screens) and removes spaces.

Currently, for whatever reason, strips the underline from the link when wrapped in a span... :shrug: 

Previous comments:
- https://github.com/nuejs/nue/commit/752e680a73079859371a3732dd0b85df2dfe0186#r144440249
- https://github.com/nuejs/nue/commit/89aaa518494962c9bdf1f83e67f7c5a7d5b2eb4e#r144876994

nobkd: Without span:

image

- breaks the text flow

---

With span:

image

- missing underline on link

tipiirai:
This was HUGE! Thank you!

PR #298, opened 2024-08-06
Update reactive-components.md

seneca:
Minor changes

### DIFF
--- a/packages/nuejs.org/docs/reactive-components.md
+++ b/packages/nuejs.org/docs/reactive-components.md
@@ -42,12 +42,12 @@ Here's the source code for the reactive component:
 </div>
 ```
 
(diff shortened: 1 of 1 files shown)


packages/nuejs.org/docs/reactive-components.md
```diff
@@ -61,7 +61,7 @@ Inline handlers are good for simple expressions.
 
 
 ### Method handlers
-More complex functionality should go to an instance method
+More complex functionality should be relegated to an instance method:
```
nobkd: I think I wouldn't understand this on a first read.
Maybe `… should be moved to …`
What do you think?

nobkd (changes requested):

packages/nuejs.org/docs/reactive-components.md
seneca: Relegated is a big word and I couldn't find a better one. 
"should be moved" is the way to go here.
Thanks.

nobkd:
```suggestion
More complex functionality should be moved to an instance method:
```
Then I'll use that. Thanks for your review.

nobkd (approved):

PR #301, opened 2024-08-06
Update tags.md

seneca:
Edits

### DIFF
--- a/packages/nuejs.org/docs/tags.md
+++ b/packages/nuejs.org/docs/tags.md
@@ -4,7 +4,7 @@ Markdown extensions or "tags" allow content authors to create rich, interactive 
 
 
 ## Tag usage
-Tags are easy to read and write. They always start on a new line, with a square bracket, followed with a tag name and options. Let’s say we want to add a video, we would write something like this:
+Tags are easy to read and write. They always start on a new line, with a square bracket, followed by a tag name and options, followed by a cloasing square bracket. Let’s say we want to add a video, we would write something like this:
 
 ``` md
 [video /videos/explainer.mp4]
@@ -35,7 +35,7 @@ Tag options can be supplied in several ways:
 ```
 
 ### Nested content
-Most tags like images, buttons, tabs, and grids accept nested content:
+Most tags like images, buttons, tabs and grids accept nested content:
 
 ``` md
 [image explainer.png]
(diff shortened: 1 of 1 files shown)


packages/nuejs.org/docs/tags.md
```diff
@@ -4,7 +4,7 @@ Markdown extensions or "tags" allow content authors to create rich, interactive
 
 
 ## Tag usage
-Tags are easy to read and write. They always start on a new line, with a square bracket, followed with a tag name and options. Let’s say we want to add a video, we would write something like this:
+Tags are easy to read and write. They always start on a new line, with a square bracket, followed by a tag name and options, followed by a cloasing square bracket. Let’s say we want to add a video, we would write something like this:
```
nobkd:
```suggestion
Tags are easy to read and write. They always start on a new line, with a square bracket, followed by a tag name and options, followed by a closing square bracket. Let’s say we want to add a video, we would write something like this:
```
typo

packages/nuejs.org/docs/tags.md
```diff
@@ -87,25 +87,25 @@ Images can link to URL's with `href` attribute:
...
+  `alt` is an alternate text for the image.
 
-  `src` image source
+  `src` image source.
 
-  `caption` image caption
+  `caption` image caption.
 
-  `href` a link URL for the image
+  `href` a URL link to the image.
```
nobkd:
```suggestion
  `href` adds a link on the image.
```

Feels like it's duplicated, because at least in my surroundings, URL and link are used interchangeably

Also, it's not a link TO the image itself (would be `src`), but to another location

I find it quite challenging to provide a short description that properly explains that it points to another URL instead of to the image itself.

Please let me know of better descriptions.

packages/nuejs.org/docs/tags.md
```diff
@@ -397,13 +397,13 @@ Render a [tabbed layout](//saadiam.medium.com/tabs-design-best-practices-8fafe93
 ```
 
 [.options]
-  #### [codetabs] options
+  #### [codetabs] options.
```
nobkd:
```suggestion
  #### [codetabs] options
```

nobkd (changes requested):

packages/nuejs.org/docs/tags.md
seneca: `href` adds a link on the image. Your change reads much better. great.

nobkd (approved):

PR #302, opened 2024-08-06
Update template-syntax.md

seneca:
Edits.

### DIFF
--- a/packages/nuejs.org/docs/template-syntax.md
+++ b/packages/nuejs.org/docs/template-syntax.md
@@ -171,7 +171,7 @@ Now we can pass custom content for the parent:
 </media-object>
 ```
 
-The slot element in the media object is replaced with the nested content in the loop. The nested content can contain anything: text, HTML tags, other custom components like product rating, commenting or product metadata.
+The slot element in the media object is replaced with the nested content in the loop. The nested content can contain anything: Text, HTML tags, other custom components like product rating, a comment section or product metadata.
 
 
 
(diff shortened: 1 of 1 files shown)


packages/nuejs.org/docs/template-syntax.md
```diff
@@ -171,7 +171,7 @@ Now we can pass custom content for the parent:
 </media-object>
 ```
 
-The slot element on the media object is replaced with the nested content on the loop. The nested content can contain anything, including text, HTML tags, and other custom components, like product rating, commenting, or product metadata.
+The slot element in the media object is replaced with the nested content in the loop. The nested content can contain anything: text, HTML tags, other custom components like product rating, commenting or product metadata.
```
nobkd:
```suggestion
The slot element in the media object is replaced with the nested content in the loop. The nested content can contain anything: Text, HTML tags, other custom components like product rating, a comment section or product metadata.
```
WDYT about using `comment section`?

packages/nuejs.org/docs/template-syntax.md
```diff
@@ -335,7 +335,7 @@ Then the final rendered button would merge both classes:
 
 
 ### Passing data with `:bind` { #bind }
-Bind directive makes every object property directly accessible on the component. Instead of writing `{ data.title }` inside the component, you can just write `{ title }`. This is particularly useful when looping components.
+Bind directive makes every object property directly accessible to the component. Instead of writing `{ data.title }` inside the component, you can just write `{ title }`. This is particularly useful when looping components.
```
nobkd:
```suggestion
The bind directive makes every object property directly accessible from the component. Instead of writing `{ data.title }` inside the component, you can just write `{ title }`. This is particularly useful when looping components.
```

Hm. I think we used different points of view. You from the parent component and me from the child component?

packages/nuejs.org/docs/template-syntax.md
```diff
@@ -568,14 +568,11 @@ You can pass the iterated data to the component with attributes:
 />
 ```
 
-Or you can use [:bind attribute](template-syntax.html#bind) to pass all the data at once:
+or you can use [:bind attribute](template-syntax.html#bind) to pass all the data at once:
```
nobkd:
```suggestion
or you can use [:bind attribute](#bind) to pass all the data at once:
```

nobkd (changes requested):

nobkd (approved):

tipiirai:
Thank you!

PR #306, opened 2024-08-07
docs: tab event key

nobkd:
Follow up to #304 

[Checked for other names of tab key](https://www.w3.org/TR/uievents-key/#named-key-attribute-values)


tipiirai:
Thanks!

PR #309, opened 2024-08-09
fix: cli create

nobkd:
Changes:
- use tarball archive (in my case a gzip compressed tarball), as `tar` CLI tool is available by default on most platforms
- support opening link on more platforms (fail silently, to keep server running, and isn't problematic if missing)
- get port from nue server
- ignore dot-files and -dirs on empty check. (helps e.g. with empty git repos)
- use node compatible file writing

Currently fails, because `https://simple-blog.nuejs.org/simple-blog.tar.gz` obviously does not exist.

> [!note]
> Feel free to reject, just had fun checking how this could work on more platforms

Future ideas:
- use directory from `-r` option, extract archive there and use it for root of serve
- maybe fix opening from wsl (I at least think, it does not work, not actually tested)

### DIFF
--- a/packages/nuekit/src/create.js
+++ b/packages/nuekit/src/create.js
@@ -1,43 +1,43 @@
-
-import { createKit } from './nuekit.js'
-import { finished } from 'node:stream/promises'
-import { createWriteStream } from 'node:fs'
 import { execSync } from 'node:child_process'
 import { promises as fs } from 'node:fs'
-import { Readable } from 'node:stream'
 
+import { createKit } from './nuekit.js'
 
 async function serve() {
   const nue = await createKit({ root: '.' })
-  await nue.serve()
+  const terminate = await nue.serve()
 
   // open welcome page
-  execSync('open http://localhost:8083/welcome/')
+  const open = process.platform == 'darwin' ? 'open' : process.platform == 'win32' ? 'start' : 'xdg-open'
+  try {
+    execSync(`${open} http://localhost:${nue.port}/welcome/`)
+  } catch {}
+  return terminate
 }
 
-export async function create({ name='simple-blog' }) {
+export async function create({ name = 'simple-blog' }) {
 
   // read files
-  const files = await fs.readdir('.')
+  const files = (await fs.readdir('.')).filter(f => !f.startsWith('.'))
 
   // already created -> serve
   if (files.includes('site.yaml')) return serve()
 
   // currently only simple-blog is available
-  if (name != 'simple-blog') return console.error('Template does not exist', name)
+  if (name != 'simple-blog') return console.error('Template does not exist:', name)
 
   // must be empty directory
-  if (files[1] || files[0] != '.DS_Store') {
-    return console.error('Please create the appplication to an empty directory')
-  }
+  if (files.length) return console.error('Please create the template to an empty directory')
 
-  // download zip
-  const zip = await fetch(`https://${ name }.nuejs.org/${name}.zip`)
-  await Bun.write('source.zip', zip)
+  // download archive
+  const archive_name = 'source.tar.gz'
+  const archive = await fetch(`https://${name}.nuejs.org/${name}.tar.gz`)
+  await fs.writeFile(archive_name, await archive.arrayBuffer())
 
-  // unzip
-  await execSync('unzip source.zip')
+  // unzip and remove archive
+  execSync(`tar -xf ${archive_name}`)
+  await fs.rm(archive_name)
 
   // serve
-  await serve()
+  return await serve()
 }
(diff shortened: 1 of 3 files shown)


packages/nuekit/src/create.js
```diff
@@ -1,43 +1,42 @@
...
+  // download archive
+  const archive_name = 'source.tar.gz'
+  const archive = await fetch(`https://${name}.nuejs.org/${name}.tar.gz`)
+  await fs.writeFile(archive_name, await archive.arrayBuffer())
 
-  // unzip
-  await execSync('unzip source.zip')
+  // unzip and remove archive
+  execSync(`tar -xf ${archive_name}`)
+  await fs.unlink(archive_name)
```
nobkd:
```suggestion
  await fs.rm(archive_name)
```

tipiirai:
WOW — I really love this update! Fixes all my concerns with the create command. I'll push a tarball there. Big thanks!

tipiirai:
I have hard time creating the tarball in osx. Using this command inside the simple-blog directory:

```
tar --exclude=.* -czvf simple-blog.tar.gz . 
```

But the generated file is empty. The file should exclude all dotfiles and dotdirs. Can you help? Thanks!

tipiirai:
I managed to create the tarball so that the contents of the file has a single "simple-blog" folder (not directly on the root). I guess that works too. 

I also created a test file `test.tar.gz` which is the same site but without the large .png and .jpg files. The file is only 10k (instead of 2M) so it's good for testing the download & unzip combo.

nobkd:
From `simple-blog`: `tar -czvf simple-blog.tar.gz *`
Creates the file in `simple-blog` and automatically exclude dot files

I tested it from bash and zsh

tipiirai:
For me "*" did not include the `@lib` and `@global` folders so we cannot use it. I'm going to change the logic so that it creates the "simple-blog" and folder and changes the working directory to there. Maybe have an option to specify the name for the new folder.

nobkd:
Okay. As I fixed the DOM tests #285 I already added the logic to use `-r` option with create command. Maybe that can just be extended for now. 

I will check, if I find a better command. Maybe even unpacking only sub directory would be possible.

I'll also let you know if I find a better command for compressing, that includes `@`-dirs et cetera.

nobkd:
May I know what shell you use? And maybe details about your tar cli version etc?

(for me * includes @ dirs but not dot dirs, as would be expected)

tipiirai:
I'm using zsh and bsdtar 3.5.3

tipiirai:
I think it makes sense to create a new directory "simple-blog" after issuing the command `nue create simple-blog`. What do you think?

nobkd:
Maybe as a default, but I personally don't find it unreasonable to already have a project name planned and unpack a starter to current directory or provide a name...
I'll think some more about it.

I guess create will be part of a later release? Or do you want to get it ready with the big release? 

---

BTW have installed bsdtar now and it says `bsdtar: Failed to open 'global'` for directory `@global` In simple-Blog with my example command, when compressing. Will check for why that is different between gnu tar and bsdtar command, and how to fix it...

Edit: I need to get some sleep now. I'll probably be active later again.

tipiirai:
I'm planning to include `nue create` for the big release, because it's simplifies the experience for new users a lot, and it's a much smaller task compared to rewrite all `create-nue` examples to be compatible with the new version. I also don't like the all the complexities / dependency issues around `bun create`.

nobkd:
I see. 

I still have to check, if it works on windows, but you can just compress the directory itself (like you already did) , and on extracting unwrap the directory:

`tar --strip-components=1 -C . -xvf simple-blog.tar.gz`

-C is the location where to extract to (can be left out if `.`), --strip-components=1 removes the simple-blog directory (or to be exact 1 level of directories I think) from the path on extracting.

Will build some code later.

tipiirai:
I just noticed the use of the `-C`, which is pretty clever. I'll ensure the create script works on my mac. I think this is a solid piece of code for the big, alpha release.

nobkd:
Just so you know, your `.tar.gz` contains a weird `._index.md` with some apple-stuff content:

image

that (I think) results in this console log:

```
tar: Ignoriere unbekanntes Schlüsselwort „LIBARCHIVE.xattr.com.apple.provenance“ für erweiterten Kopfteil
```

I think, that's just a temporary file from Apple, (that maybe got wrongly encoded on packaging)

¯\\\_(ツ)\_/¯

nobkd:
Just found out, that you can use git to compress stuff, that is commited:
```
git archive -o test.tar.gz HEAD:packages/examples/simple-blog/
```

run from the `root` of the project, includes only contents of simple-blog (at root of dir, but can be added to subdir instead with `--prefix dirname/` ), that are committed (at least locally)

This is useful e.g. to automatically exclude generated data e.g. with `nue build` or fiies automatically added by the filesystem (see comment above).#

Ref: https://git-scm.com/docs/git-archive

PR #310, opened 2024-08-10
fix: replace cr and crlf with lf, split on all line endings

### LINKED ISSUE #307: Simple Blog Post cannot load hero image & "Invalid date"?
Hi. I was trying to build a simple blog with `bun create nue@latest` command but when I add another "blog-name.md" into the "posts" section, hero image of the post never shows up and date cannot be displayed. Its content appears but hero & date never works. Here's a preview:

hero

The content are exactly the same (its from the example blog posts I'm using that comes with Nue) with different names, one fails to work which is the file I created.

I'm in a Windows environment, running Nue in WSL2.

nobkd:
- fixes #307 

I tried to replace / split as early as needed, to keep the replace / split changes small.
I also tried checking for places where functions get exported.
Changed in all packages for standalone use.

Added a test, that might not catch everything, but that's probably okay.

The change in `packages/nuekit/src/builder.js` will probably be different when #278 is merged.


tipiirai:
Looks good. Maybe the complex line ending could be a constant?

nobkd:
Yeah, maybe. But would have to be per package to prevent unnecessary dependencies.

tipiirai:
Good point. This is better.

PR #311, opened 2024-08-12
chore(deps): upgrade

nobkd:
I still need to update some code involving marked having a new renderer, and amybe some other stuff.

See: https://github.com/markedjs/marked/releases/tag/v13.0.0
And: https://github.com/markedjs/marked/releases/tag/v14.0.0

The tests might not fail, but I think, there's definitely still something wrong, at least with heading rendering


nobkd:
I probably did something wrong, so please check

tipiirai:
Looks solid. Thanks a lot!

nobkd:
#312 will downgrade marked for now, as heading rendering is still wrong.

tipiirai:
Cool. In general, I'm not a fan of Marked since there is no way to alter the AST before it gets rendered. Also the API is a bit of a mess. Not sure what is better though. One option is to implement a fully customized Nuemark parser

PR #312, opened 2024-08-13
fix: downgrade marked (important)

nobkd:
Fixes heading stuff

see:
- https://github.com/nuejs/nue/commit/ea51b2598d5a2e6267ebd7560e088538807db878#commitcomment-145318757
- https://github.com/nuejs/nue/commit/ea51b2598d5a2e6267ebd7560e088538807db878#commitcomment-145329712

---

Downgrading is currently easier than fixing. Will try again in the future, but I'll probably not do it before you want to do a release.


tipiirai:
Thanks! Better this way. Not sure if Marked will be part of Nue's long-term plan. Definitely looking for options.

PR #316, opened 2024-08-14
Nue 1.0 (beta)

tipiirai:
A major release


tipiirai:
@nobkd as expected, the dev branch has conflicts with the master. Do you know if there is anything worth taking from the master? Do you have experience in merging these big changes? Is there an easy way to do a force-merge without conflicts and then maybe cherry pick only what is needed. Thanks!

nobkd:
I'm not sure.

Here are the commits you did to master after you moved most commits to dev: https://github.com/nuejs/nue/compare/dev...master

These should be the only changes you did, so maybe you can just remove them from master and redo them, when they are important?

---

That's just 3 commits:

- https://github.com/nuejs/nue/commit/0813c6c44e0c0a40a2cedb6e25e4413beb7042ad (skip long header id's. Is the largest commit, and I don't know if it is already done differently on dev)
- https://github.com/nuejs/nue/commit/9338c807c1653fbde7b947da198da1c143323a8a (monster update notice, can probably removed)
- https://github.com/nuejs/nue/commit/043ce76236e2e3365f4d7893ba956e2affba6161 (small readme changes)

---

Of course you can also just try a local merge if you want to, and decide for/against changes in a merge editor. Will probably be the easiest, if the changes aren't too huge... 🤷‍♂️ 

---

You can also create a new branch based on up to date dev, merge master in the new branch, run tests, etc. and then merge that new branch into dev. then the merge conflicts between master and dev should be resolved.

tipiirai:
True. There are far less conflicts than I thought. I'll probably just use the web editor to resolve them, which is my favourite Git client :)

tipiirai:
btw: does Nue work on Windows?

nobkd:
> btw: does Nue work on Windows?

It should. The PRs that fixed the path issues, have been merged looong ago, there just was no release after that.

The old `create-nue` package still fails because of nuejs/create-nue#43 not being merged+released, but as we now have a new command that should be irrelevant.

I can still do some tests now, to confirm that it works. (I'll start my vm now)

btw: when I tested stuff with bun on windows, there were no issues. so bun seems to be stable there.

nobkd:
Found no issues with nue on windows for now. I'll try to find a way to start nue with node, but where.exe doesn't work as i would expect, and that there are extra executables on windows doesn't make it better...

---

Okay... when running `nue -r .\simple-blog\` (auto-completed), I get a problem real fast.....

`\.dist\dev\.dist\dev\.dist\dev\.dist\dev\.dist\dev\.dist\dev\.dist\dev\.dist\dev\.dist\dev\.dist\dev\.dist\dev\.dist\dev\.dist\dev\.dist\dev\.dist\dev\.dist\dev\.dist\dev\.dist\dev\.dist\dev\img\favicon.jpg`

not sure what happened there, but I'll investigate that sooner or later.. okay. seems like somewhere there are issues with `\`es... running with just `nue -r simple-blog\` also triggers the issue. `nue -r Desktop\simple-blog` and `.\Desktop\simple-blog` doesn't, so it has to have to do with ending with backslash.

Will try to find the reason tomorrow, but can't predict to really resolve it :sweat_smile:

---

running create works flawlessly, building nuejs.org as well, page-router test project also works.

Only problem for now stays the backslashes on end of root

packages/nuekit/src/init.js
```diff
@@ -1,22 +1,21 @@
...
   const cwd = process.cwd()
-  const srcdir = getSourceDir()
   const outdir = join(cwd, dist, '@nue')
 
-
   // has all latest?
-  const latest = join(outdir, '.043')
+  const latest = join(outdir, '.05')
   try {
     if (force) doError()
```
nobkd, line 20:
```suggestion
    if (force) throw new Error()
```

I find this is more readable. `doError()` could be a function. this not.
but it's your decision.

tipiirai: Agreed. `throw new Error()` reads better

tipiirai:
Sounds very promising! Looks like we can get Windows support for the official 1.0

PR #317, opened 2024-08-14
docs: fixes

nobkd:


packages/nuejs.org/docs/installation.md
```diff
@@ -24,48 +24,44 @@ Create your first Nue website:
 
 
 ```sh
-# create an empty directory and cd to it
-mkdir "simple-blog" && cd "$_"
```
tipiirai, line 28: No mkdir step needed?

nobkd: Is included in create:

https://github.com/nuejs/nue/blob/e2119f3821764184745a56363acb43f4c77f18c4/packages/nuekit/src/create.js#L37

Is used to prevent different `root`s from failing.

(On a side note: `cd $_` does not work on windows, `&&` also not. I've tested it a few days ago)

---

created directory defaults to project starter template name:

https://github.com/nuejs/nue/blob/e2119f3821764184745a56363acb43f4c77f18c4/packages/nuekit/src/create.js#L17-L18
(name is always overridden, default in function definitin can be removed)

packages/nuejs.org/site.yaml
```diff
@@ -6,7 +6,7 @@ libs: ["@lib", docs/lib, blog/lib]
 og: /img/og-blue-big.png
 view_transitions: true
 native_css_nesting: true
-origin: https://www2.nuejs.org
+origin: https://www2.nuejs.org # TODO: update to correct host on release
```
nobkd, line 9: I can also already change it, but I don't know when you would move to the main domain.

packages/nuejs.org/docs/installation.md
tipiirai: Awesome. In what case(s) `mkdir` gets called?

The root defaults to "."

https://github.com/nuejs/nue/blob/e2119f3821764184745a56363acb43f4c77f18c4/packages/nuekit/src/cli.js#L85

nobkd: nope. changed that (commit: https://github.com/nuejs/nue/commit/9b1dd82352f667e441fdb12eb94653c8eb86f96c). the create command gets the `root` variable, not the `args.root` one that is used by everything else. `root` is `null` if not user-defined via `-r` option.

`mkdir` gets called, when `root` does not exist. see the `if` to the `else` from before:

https://github.com/nuejs/nue/blob/e2119f3821764184745a56363acb43f4c77f18c4/packages/nuekit/src/create.js#L27
-> check if empty, etc.

(maybe have to check, if it is a directory too...)

tipiirai: I see, but in what scenarios the root does not exist when I'm using the command line?

tipiirai: I guess the site gets created to a new directory called `simple-blog` unless explicitly specified with `nue create --root my-website`?

packages/nuejs.org/site.yaml
tipiirai: Good point. I'll change that tomorrow.

packages/nuejs.org/docs/installation.md
nobkd: yes. and if the directory already exists, it checks, whether it is empty.

nobkd: > I see, but in what scenarios the root does not exist when I'm using the command line?

If it hasn't been created by hand beforehand, it gets created, to prevent from just failing, which would be unexpected.

tipiirai: Cool. I'll add a note on this to the installation doc. Merging this now.

tipiirai:
Thanks!!

PR #326, opened 2024-08-18
fix: svg className readonly

### LINKED ISSUE #151: SVG reports an error as a component
**SVG component definition:**
```
<svg @name="sun-svg" class="{class}" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24">
  <path />
</svg> 
```
**calling component:**
`<sun-svg class="swap-on"></sun-svg>`

**The className of svg is a read-only attribute:**
image

nobkd:
- resolve #151

Tested in personal project with (works) and without (fails) the change.

---

other changes:
- unescape some test cases by using `'` over `"`


tipiirai:
WOW! A Nue JS fix 🥇  Thanks!

PR #328, opened 2024-08-19
Update tutorial.md

seneca:
"nue create simple-blog" creates a folders structure with `welcome` folder and not `home` folder.
The reader of the tutorial may be confused not finding the "home" folder in the generated folder structure.

### DIFF
--- a/packages/nuejs.org/docs/tutorial.md
+++ b/packages/nuejs.org/docs/tutorial.md
@@ -38,7 +38,7 @@ Here's how our blog is structured:
 
 [.folders]
   - `index.md` the front page
-  - `home` front page styling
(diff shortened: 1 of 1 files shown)


nobkd:
I understand what you mean, and the folder structure differs, but that's not what was meant there.

The `home` directory is used to contain components / styles for only the home / front page.
The `welcome` directory can be left out of the docs, because it only has the purpose of welcoming the person, who created the project.

It's probably best, to add some front page only styles to the `home` directory, to fix that tutorial-project-inconsistency.

tipiirai:
Thank you! 

ps: I was touring Europe for ten days, but now I'm back coding.

PR #329, opened 2024-08-19
chore: formatting

nobkd:
- formatting
- mostly whitespace changes
- some typo fixes


tipiirai:
Wow. This is a major update. I tend to use extra new lines between things due to aesthetic reasons. Do you prefer a tighter formatting?

nobkd:
I usually have only one newline to split ideas / functions / ..., this allows me to see more code in the short area of focus, while keeping a break between ideas / functions / ...

In e. g. Markdown lists I don't use extra lines, because they are separated through the `*`, `-` or `1.` indicator....

In the end, you can still decide if you like it or not. And I theoretically can also undo some changes.

packages/nuekit/src/cli.js
```diff
@@ -86,7 +86,7 @@ async function printVersion() {
 
 async function runCommand(args) {
   const { createKit } = await import('./nuekit.js')
-  const { cmd='serve', dryrun, deploy, root=null, port } = args
+  const { cmd='serve', dryrun, deploy, root = null, port } = args
```
nobkd:
```suggestion
  const { cmd = 'serve', dryrun, deploy, root = null, port } = args
```

nobkd:
Closing, and maybe later reopening with smaller scope.

PR #331, opened 2024-08-20
chore: simplifications, cleanup

nobkd:


packages/nuekit/src/builder.js
```diff
@@ -7,9 +7,8 @@ import { resolve } from 'import-meta-resolve'
 
 
 export async function getBuilder(is_esbuild) {
-  const actual_cwd = process.env.ACTUAL_CWD || process.cwd()
   try {
-    return is_esbuild ? await import(resolve('esbuild', `file://${actual_cwd}/`)) : Bun
+    return is_esbuild ? await import(resolve('esbuild', import.meta.url)) : Bun
```
nobkd: Will have to check, if this is really fine, like this...

Edit: nope, wasn't. Should work now.

tipiirai:
Looks great!

Issue #282, opened 2024-06-29
Option for pretty (extension-less) routes


nobkd:
We have a file `<project-root>/x.md` which will currently get transformed to `/x.html` on the web.

Allow transformation to `/x` or `/x/` by putting the resulting file to `/x/index.html`

korabdoroci:
Wouldn't it be simpler to just have a tag in the frontmatter instead?

nobkd:
Hmm. But what if I want that behavior for all pages, not just select ones?
Then I would have to add the resulting file path in the front matter of each file.
I can understand the pros of that approach, but that would be great to have additionally to the option for pretty links (where I don't have to care about the file extension of pages)

(it'll probably be hard enough to keep the paths to the assets of a page correct)

korabdoroci:
I was thinking something like: "pretty_routes: true" in the site.yaml which would then apply to the whole site. You could also do it in each app.yaml.

nobkd:
I was actually thinking the same and wanted a config option for `site.yaml`.
But I probably misunderstood, because of mentioning front-matter (which is `yaml` config at the start of a `md` page to my knowledge)

korabdoroci:
Yes, I was unclear. I think this would be a great addition to Nue.

PR #334, opened 2024-08-29
fix: handle unknown mime

nobkd:
If mime type not in default mime type object, let browser determine how to handle response.

Fix navigation to other file types than HTML (or extensionless (base) route), when using view transitions


tipiirai:
Looks good. At some point we want a configuration option for pretty URL's so that `blog/foo.md` translates to `blog/foo`

nobkd:
Yes. I have an issue for that: #282

PR #335, opened 2024-08-31
fix!: make components case-insensitive

### LINKED ISSUE #100: child component      name    Contains uppercase letters  ，error  not show   、ref  undefined

nobkd:
This is just some idea I had when reading #100, and it would fix #100, but I'm not sure, if this is a good idea. This just makes components case-insensitive and passes the access to the object through a proxy that makes all key accesses lower-case. This PR solves the problem, because **HTML elements are case-insensitive for the browser by default**, but currently might allow collisions on Component creation...

Well... This is just a proposal to make components case-insensitive in the end...
Okay, I'm really repeating myself.

This was just a quick sketch to outline the idea and might have some obvious flaws, and the code can be changed more, to make this better integrated (and fix some possible issues)...

Feel free to reject and let me know your opinion about this.

### DIFF
--- a/packages/nuejs/src/browser/nue.js
+++ b/packages/nuejs/src/browser/nue.js
@@ -213,7 +213,10 @@ export default function createApp(component, data={}, deps=[], $parent={}) {
 
   // context
   let impl = {}
-  const CaseInsensitiveObject = { get: function (target, name) { return target[name?.toLowerCase()] } }
+  const CaseInsensitiveObject = {
+    get(target, key) { return target[key?.toLowerCase()] },
+    set(target, key, val) { target[key?.toLowerCase()] = val },
+  }
 
   const self = {
     update,
--- a/packages/nuejs/src/browser/nue.js
+++ b/packages/nuejs/src/browser/nue.js
@@ -213,6 +213,7 @@ export default function createApp(component, data={}, deps=[], $parent={}) {
 
   // context
   let impl = {}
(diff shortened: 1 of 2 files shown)


tipiirai:
I think this is a good change. Makes sense to be compliant with the HTML specs. I did not merge this yet since you were concerned about the potential flaws. I guess tests are needed. Do you think proxy is the easiest way to do this?

nobkd:
To be honest, this is the first time I really used Proxy (because it seemed like a really easy way, to get case-insensitive access to an object). So I'm really not sure, if this is the way to go. I will check, if I see an easier / better way, and then try adding tests.

But I'm currently a bit preoccupied with personal matters, so this may (or may not) take a while.

Feel free to take this over, if you have a good idea, or want this included fast.

tipiirai:
Take your time on this. I'm not in a hurry with Nue JS issues as they are all delayed for later

nobkd:
I currently don't want to work on this, if someone wants to take this over, you can redo this on your end / copy my cranch / commits over from my fork

PR #337, opened 2024-09-01
fix: allow textarea resizing only vertically

nobkd:
prevents resizing text area outside of page creating horizontal overflow


tipiirai:
Thanks!

PR #338, opened 2024-09-01
chore: exclude examples and docs from triggering test run

nobkd:
https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore


tipiirai:
Good change

PR #341, opened 2024-09-03
test: more readable matrix

nobkd:
The matrix tries all combinations of its keys
The exclude key removes options from the resulting cartesian product

```yaml
key1: [A, B, C]
key2: [x, y]

result: [Ax, Ay, Bx, By, Cx, Cy]
```


tipiirai:
Great!

PR #342, opened 2024-09-04
feat: more `og` data, simpler `meta property` addition

nobkd:
Problem with og elements currently is, that they are not replaced on `view_transitions`.
But typically you wouldn't care about that, because you don't see them from the page anyways, and instead when you share e.g. a link (which is a fresh load).


tipiirai:
I'm currently rewriting the lousy dom diffing algorithm on the view-transitions.js module into something more generic. Not trivial, but doable.

PR #347, opened 2024-09-10
chore: async exists

nobkd:
Follow up to https://github.com/nuejs/nue/commit/6b0181e12319767f36cf0e876784bec3fab0abef

> [!Note]
> Updates to `dev` branch

You seem to have broken some tests. (Not so sure about the `create` one though)

Should I change the test-run workflow to include `dev` branch, so you see that something lets the tests fail, sooner?

### DIFF
--- a/packages/nuekit/src/nuekit.js
+++ b/packages/nuekit/src/nuekit.js
@@ -31,7 +31,7 @@ export async function createKit(args) {
   }
 
 
-  if (!(await fs.exists('site.yaml'))) {
+  if (!await fs.exists('site.yaml')) {
     console.error('No site.yaml found. Please go to project root\n')
     return false
   }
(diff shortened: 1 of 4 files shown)


packages/nuekit/src/nuekit.js
```diff
@@ -31,7 +31,7 @@ export async function createKit(args) {
   }
 
 
-  if (!(await fs.exists('site.yaml'))) {
+  if (!await fs.exists('site.yaml')) {
```
tipiirai, line 34: Cool. Didn't know the brackets can be removed here. Much cleaner.

tipiirai:
I'll fix the broken tests

tipiirai:
No need to include `dev` branch for tests. I usually fix them in one bunch after doing several updates to the codebase. Probably a bad habit, but that's how I work currently.

PR #354, opened 2024-09-12
chore: exclude tests from npm

nobkd:
Excludes tests from files to be published to npm.

(commands run from `packages` sub-dir)

Package sizes in code (with tests):
```sh
> du -hcs nuejs glow nuekit nuemark                
140K    nuejs
112K    glow
196K    nuekit
76K     nuemark
524K    total
```

Sizes of only tests:
```sh
> du -hcs nuejs/test glow/test nuekit/test nuemark/test
72K     nuejs/test
64K     glow/test
68K     nuekit/test
28K     nuemark/test
232K    total
```

232K/524K: About 44% of nue's code size on npm is tests, that gets shipped to every user, but probably isn't used by a single one.

I think it would be fine to just exclude them from npm, but in the end, it's your decision.


tipiirai:
This is good. Thanks

PR #355, opened 2024-09-12
Beta 2

tipiirai:
Some new features + a lot of bug fixes. Check commit history.


tipiirai:
@nobkd any idea why `await fs.exists()` does not work in node/windows?

goblinfactory:
> @nobkd any idea why `await fs.exists()` does not work in node/windows?

I'm not a ts dev; but Phind AI tells me that fs.exists() is deprecated, and it referenced a few links; here's the one link (discussion) I think containing the most relevant information. 
https://stackoverflow.com/questions/50767829/why-node-js-fs-existssync-doesnt-work-well-when-wrapped-in-promise

nobkd:
Were using `fs.promises.exists()` which I can't find any documentation for. Probably only implemented by bun?
And the callback version `fs.exists()` is deprecated and not useful for our use case, so we'll have to use `fs.existsSync()`.

Basically reverting my pr #347 and replace your usage with sync too.

tipiirai:
Okay. Gotcha. Oddly it's implemented in Bun, but not in Node. I'll convert the codebase to the sync version.

nobkd:
No `nuemark` version bump?

PS: don't forget replacing `*` with `latest` on release to npm

PR #365, opened 2024-09-16
lint: format js

nobkd:
Most changes are formatting or whitespace.

Imports are now categorized in 3 categories and ordered by alphabet respectively:

- `node:*` internals
- package imports
- internal imports / imports from the current package

After imports and top level variable definitions, there are two newlines separating them from the actual code. 
Exactly one newline at the end of the js file.


packages/nuekit/src/layout/head.js
```diff
@@ -76,7 +76,7 @@ export function renderHead(data) {
     const css = inline_css.map(el => el.css).join('')
     head.push(elem('style', css))
 
-  // dev mode: keep path info for hot-reloading
+    // dev mode: keep path info for hot-reloading
```
nobkd:
```suggestion
  // dev mode: keep path info for hot-reloading
```

packages/glow/src/glow.js
```diff
@@ -166,7 +166,7 @@ export function parseRow(row, lang) {
   for (const el of tags) {
     const { re, shift } = el
 
-    row.replace(re, function(match, start, n) {
+    row.replace(re, function (match, start, n) {
```
tipiirai: I personally prefer not to have a whitespace here. Or is this some sort of standard?

packages/glow/src/glow.js
```diff
@@ -263,7 +263,7 @@ export function parseSyntax(str, lang) {
 
 
 // code, { language: 'js', numbered: true }
-export function glow(str, opts={}) {
+export function glow(str, opts = {}) {
```
tipiirai, line 266: No whitespace after the method name here. Was this a slip?

tipiirai:
I'm good with the changes, except for the whitespace prior function name — which I'd like to discuss more

nobkd:
This part was automatically formatted, but I also don't like the space there, so I changed it.

PR #367, opened 2024-09-18
minor textual change

seneca:
I'm not sure whether the author wanted to say that Nue is in a completely different ballpark _than_ what you see with Rust, etc., or whether Nue is in the ballpark _of_ Rust, etc., which would mean in the same ballpark (my change here).

### DIFF
--- a/packages/nuejs.org/docs/content.md
+++ b/packages/nuejs.org/docs/content.md
@@ -119,7 +119,7 @@ Again, the name "stack" must be implemented in your website CSS by the UX develo
 
 
 ### Grids
-[Grid](tags.html#grids) is a built-in tag used like the stack but is meant for more complex layouts. They have more configuration options and their visual appearance and behavior can be enhanced with JavaScript.
+[Grid](tags.html#grids) is a built-in tag used like the stack but is meant for more complex layouts. They have more configuration options for their visual appearance and behavior can be enhanced with JavaScript.
 
 
 ```md
--- a/packages/nuejs.org/docs/content.md
+++ b/packages/nuejs.org/docs/content.md
@@ -62,7 +62,7 @@ og: /img/hero-image.png
 
 
 ### Sections
-Just like books can be divided into chapters, your long-form articles and landing pages are often divided into sections. These sections are separated with three or more dash (`---`) or equals (`===`) characters:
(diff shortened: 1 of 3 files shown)


packages/nuejs.org/docs/tutorial.md
```diff
@@ -348,10 +348,6 @@ The production-optimized version is generated with the `nue build --production`
   large: /img/blog-generate-big.png
   width: 500
 
-This step is extremely fast. No matter how large your website, the generation times are in the ballpark than what you see with **Rust** and **Go** -based generators like **Hugo**.
+This step is extremely fast. No matter how large your website, the generation times are in the ballpark of what you see with **Rust** and **Go** -based generators like **Hugo**.
```
nobkd:
```suggestion
This step is extremely fast. No matter how large your website, the generation times are in the ballpark of what you see with **Rust** and **Go**-based generators like **Hugo**.
```

nobkd (changes requested):

nobkd (approved):

nobkd:
Your change sounds correct to me, but I'll let `tipiirai` look over it too.

packages/nuejs.org/docs/content.md
```diff
@@ -119,7 +119,7 @@ Again, the name "stack" must be implemented in your website CSS by the UX develo
 
 
 ### Grids
-[Grid](tags.html#grids) is a built-in tag used like the stack but is meant for more complex layouts. They have more configuration options and their visual appearance and behavior can be enhanced with JavaScript.
+[Grid](tags.html#grids) is a built-in tag used like the stack but is meant for more complex layouts. They have more configuration options for their visual appearance and behavior can be enhanced with JavaScript.
```
nobkd, line 122: I don't think this sentence is understandable (and the change to `for` changes the meaning and breaks grammar (see `and behavior` missing `their` in between))

nobkd (changes requested):

packages/nuejs.org/docs/content.md
seneca: I hope the author will be able to look into this.

tipiirai: Makes sense. Merging this and pushing live

PR #371, opened 2024-09-18
feat: try bun then node

nobkd:
This is just a concept idea, and was only tested on my Linux machine for now


tipiirai:
Looks great! Would simplify the installation docs a lot

nobkd:
This behavior (of first using bun, if bun not found, using node) would still have to be mentioned in the docs.

note: I'll now test if this works on my windows vm, (using `bun link` from `nuekit` dir) to see, if this shebang is even used on windows, or if this depends on the package manager with which the package was installed.

Update:
Doesn't work on windows, because it fails to find the `/bin/sh` executable... Not sure, if this is possible, then.

nobkd:
I don't see a way forward now (with Windows), but might revisit this in the future, if I find a clue, on how to proceed.

Closing for now (sadly)

PR #372, opened 2024-09-19
fix: block ext spacing

nobkd:
Ref #368

Only allows extra spaces at end of line

(Not sure  if this is still needed, when you're done with your markdown parser)


tipiirai:
Thanks.

btw. Nuemark 2.0-beta coming up soon, which is a full rewrite and runs standalone without the marked dependency

PR #373, opened 2024-09-24
docs: link to edit docs page on gh

nobkd:
Adds a link to the md file on gh, to allow editing it without searching for it in the repos.

Feel free to replace the icon e.g. with a pencil icon.

Current look (with url example on hover):
grafik

(I tried adding a margin on top, to split the link more from the content, but it was collapsed, and I didn't look any further.)


tipiirai:
Looks good. I'll edit this a bit before pushing. Thanks

Issue #369, opened 2024-09-18
`<slot for="content" />` doesn't work


nobkd:
### Describe the Bug

`<slot for="content" />` is ignored and does not get replaced with the rendered md.

### Environment

- OS: Fedora Linux 40
- Nue 1.0.0-beta.2 • Bun 1.1.27
 
### Minimal Reproduction

`layout.html`
```html
<main>
  <slot for="content" />
</main>
```

`index.md`
```md
# title
```

---

result:

```html
<main></main>
```

expected:

```html
<main>
  <h1>title</h1>
</main>
```

tipiirai:
Probably works currently with `<slot for="layout.article" />`. 

Thinking of changing the name to `<slot for="layout.content" />` and updating the docs

nobkd:
Closing, as docs are fixed on dev

PR #374, opened 2024-09-24
Update custom-layouts.md

seneca:
In the picture above the blue box containing aside, page head, page foot and beside is not named. Shouldn't it be named main?

### DIFF
--- a/packages/nuejs.org/docs/custom-layouts.md
+++ b/packages/nuejs.org/docs/custom-layouts.md
@@ -5,6 +5,7 @@ The design system consists of various "slots" that you can fill or replace with 
 [image.bordered /img/custom-layout.svg size="569 × 634 px"]
 
 
+
 For example, if you want to add a custom banner above the global header you'd create a layout module called "banner":
 
 ```html


nobkd:
If you mean this image, I agree, that the main area should be named `main` (or with the slot name? See https://github.com/nuejs/nue/issues/369#issuecomment-2370652249)


from here: https://nuejs.org/docs/custom-layouts.html

tomByrer:
@nobkd is the right hand side label `beside` a typo?
Seems it also should be `aside`, or maybe `aside @complementary`?
https://github.com/nuejs/nue/blob/master/packages/nuejs.org/docs/layout.html#L16

nobkd:
No, iirc it isn't. All bold values are html `elements` (with slot I think), all non-bold ones are `@name="<component>"` layout components, you can define in your layout component files (maybe it can even be a non-layout component, but I'm not sure, as I haven't tested that one)
For the components, you can decide, what the html element is, yourself

You can see an usage example for pagefoot in #373

tipiirai:
I'm doing a significant update to the documentation. Make it much clearer and the product should feel much less "magical". Here is a new picture for this specific topic I'm going to use

nobkd:
Wasn't `<footer>` something different from `@name="pagefoot"`?
Imo it makes sense to have a footer element at the bottom of the page, not inside the main at all, but e.g. where you put `@bottom` in your graphic. Maybe even just replacing it.
I'd have to recheck how it is semantically, where you put the footer element, but I always thought, you should use one footer, at the end of the page, not inside main

PR #375, opened 2024-09-26
fix(nuemark2): line splitting, import, undefined blocks

nobkd:
This `EOL` from `node:os` does not work for splitting a file on line-endings, because `EOL` is `\n` on Mac/Linux and `\r\n` on Windows. BUT files ´on Windows can also just use `\n` line endings, as Windows can also handle those.

The package from npm is named `nue-glow` not `glow` (and should also have this name in your local installation)

When I ran `bun test` the `block` variable was undefined and some test errored on unpacking.

(There are still routing tests failing, idk why, but I currently don't care about them.)

### DIFF
--- a/packages/nuemark2/index.js
+++ b/packages/nuemark2/index.js
@@ -1,8 +1,8 @@
 
 import { renderLines } from './src/render-blocks.js'
 import { parseDocument } from './src/document.js'
-import { EOL } from 'node:os'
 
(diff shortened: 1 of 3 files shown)


tipiirai:
Thanks!

tipiirai:
I'm not sure about those routing tests whether they do the best job at finding issues: slow, complex and error-prone. I'd like to find a simpler alternative.

PR #376, opened 2024-09-27
test(dev): fix router test

nobkd:
This fixes the router test (mentioned in #375)

> [!Note]
> Problem now is, that you can't identify what a component was, when it is rendered, from the DOM alone.
> What do you think about maybe adding `custom="<name>"` instead of only `custom`?

---

PS: (to https://github.com/nuejs/nue/pull/375#issuecomment-2378462550)
these tests are only slow when they time out.
I agree, that they're maybe complex and error-prone, but that's also, because the API changes.
I hope, that I will take some time to refactor them soon, so they get less complex and error-prone. (Maybe even adding more tests to ensure DOM gets changed as expected.)

### DIFF
--- a/packages/nuekit/package.json
+++ b/packages/nuekit/package.json
@@ -17,9 +17,9 @@
     "bun": ">= 1",
     "node": ">= 18"
   },
-	"scripts": {
-		"test": "node --experimental-vm-modules ../../node_modules/jest/bin/jest.js --runInBand"
-	},
+  "scripts": {
+    "test": "node --experimental-vm-modules ../../node_modules/jest/bin/jest.js --runInBand"
+  },
   "dependencies": {
     "diff-dom": "^5.1.4",
     "es-main": "^1.3.0",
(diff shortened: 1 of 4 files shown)


nobkd:
OK, changed the code, so the component keeps the name. I know, that the name is duplicated now, but I at least still know what the component was, once rendered.

Just undo the last commit, if you don't agree with this change.

tipiirai:
Thanks. 

Is there a specific reason for keeping the name value? Does it help with testing? (You can get the name via element.tagName too)

nobkd:
@tipiirai 

Of course, this isn't always needed.
But see the following, when running source linked (`bun link` in `packages/nuekit`) `nue` from `packages/nuekit/test/page-router-test` before my change (adding the `name` to `custom`):

Before nue renders components (rendered index.html):

```html
<main>
  <article>
    <h1>Home</h1>
    <app custom></app>
  </article>
</main>
```

After nue rendered components (opening page in browser):

```html
<main>
  <article>
    <h1>Home</h1>
    <section custom>
      <h2>App mounted</h2>
    </section>
  </article>
</main>
```

So... how would you now know what this component ever was, when you don't have access to nue's in memory component structure?
`element.tagName` gives `section`, if you just select the `[custom]` elements.
You might never need this from external, but at least testing DOM gets much easier when adding the `name` to the element somehow! (especially when there are multiple components in the page)

PR #378, opened 2024-09-28
fix(dev): keep comp name

nobkd:
Follow up to #376, to also render component `name` to `custom` attribute with `nuejs-core`.

My reasoning: https://github.com/nuejs/nue/pull/376#issuecomment-2380656300

Still, your decision, if you accept this change, or ignore this PR and undo the last commit from #376


tipiirai:
Thanks. The reasoning made perfect sense

PR #383, opened 2024-10-11
test: run tests only on main repo / manual dispatch

nobkd:
Tests get skipped on forks, manual test runs still work on forks.


nobkd:
Change should not affect anything else, so merging now.

PR #384, opened 2024-10-11
feat(glow): option to disable prefixes / mark

nobkd:
Adds ability to just highlight code without worrying to lose some data.
You can just throw in your code, and be sure, that no characters are removed, and just highlighting happens.

Might not be that interesting for nuekit / nuemark, but for use outside of the nue-ecosystem.

[Example](https://nobkd.github.io/nue-editor/): [code change](https://github.com/nobkd/nue-editor/commit/d344256046630cb025d727f2958de79086ababea)

Will probably have to add docs to the readme about these options.

### DIFF
--- a/packages/glow/src/glow.js
+++ b/packages/glow/src/glow.js
@@ -179,14 +179,14 @@ export function parseRow(row, lang) {
 }
 
 function renderString(str) {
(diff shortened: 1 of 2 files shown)


tipiirai:
makes sense. thanks

PR #385, opened 2024-10-19
test: check docs and READMEs links

nobkd:


nobkd:
I might extend this in the future, to also check links in the `examples`, but I think the docs are the most important right now, to keep all links fresh and prevent broken links.

PS: this will probably fail on the docs in `dev` right now, because many paths are in change.
This should not even run on dev if not run manually.

PR #387, opened 2024-10-22
perf: keep js builder

nobkd:
keep js builder instead of re-resolving all the time.

I might need to change some code, so the esbuild tests really use the esbuild builder...
I don't know if it really matters, because it still get's tested on the jest+node test


tipiirai:
Makes sense

PR #388, opened 2024-10-22
feat(nuemark): known inline `<mark>` syntax

nobkd:
TODO:
- [ ] test


nobkd:
OK. Closing this. I somehow thought the `|` syntax was for whole lines only, but according to your updated docs it works for `|inline|` too.

Thought maybe (@tipiirai) you're still interested, because the `<mark>` syntax with `==marked==` is much more common in my opinion: https://www.markdownguide.org/extended-syntax/#highlight


(i also thought strikethrough would be with two tilde's on both ends (`~~strike through~~`)

PR #391, opened 2024-10-26
fix: email format and send email functioning in Nue FAQ

kappalasaimohith:
Fixed email display in FAQ to ensure it shows correctly and is clickable.


kappalasaimohith:
Fixed email display in FAQ to ensure it shows correctly and is clickable.

nobkd:
I think this was intentional, so bots don't scrape it as easily.
@tipiirai can probably tell you better than me, because it's just an assumption on my end :)

tipiirai:
Yes. This was intentional. Not sure if this a valid trick anymore. I bet that today's bots can scrape that. What do you think?

kappalasaimohith:
@tipiirai and @nobkd  Thanks for the clarification! I understand it was intentional, and I agree it may not hold up against modern bots.

PR #395, opened 2024-11-02
feat: use "workspace:" protocol for easier publish

nobkd:
Follow up to #261

Pro:
- Automatically replaces `workspace:*` version with the actual package version from registry (!) (at least it seemed so in my test, soo... publish order matters) on `bun pm pack` / `bun publish` ([docs](https://bun.sh/docs/cli/publish))
- no more manual `latest` replace for `*`-version

Con:
- Needs patch *before* install for local `npm` development setup. See `setup-npm.js`.

---

Disclaimer: I've never used `bun publish` and only tested `bun pm pack` on `nuekit` package and looked inside the packed `package.json` where the version got replaced just fine


tipiirai:
So what are the steps when publishing a new version to NPM? Is there a revert script to go back to the default package.json format?

nobkd:
1. Should afaik work the same as with `npm` but using the `bun` command instead.
2. See `setup-npm.js`


To 1.: I'm not sure, if there might arise issues regarding a **fixed version** (instead of a version range, I need to check, if that's possible) through replacement on `bun publish`, but if we always release all libraries, that depend on another one from our workspace, there should be no issue.

I might edit this comment tommorrow again, and clarify my text some more, since I'm currently a bit low on time :(

nobkd:
closing for now, i might reopen later. I really think, this should be done, because currently all packages have `latest` as version, and if a older version of `nuekit` is used and the `latest` version of e.g. `nuemark` the api between them might differ. so I think having a version constraint makes sense. But there are probably also cons to this approach here

PR #397, opened 2024-11-08
test: extend link checks

nobkd:


tipiirai:
This would indeed be a great test for the website

nobkd:
I'm still looking, if I find better options to make this code more readable and maybe even shorter.

nobkd:
Okay, I think I'm done for now. I can't simplify it (at least currently).

I fixed a few links, but on most of the ones I left, I'm unsure, where exactly I should link them to. I don't know the new docs enough to be sure about this.

Here you can see the link check output rendered from md: <https://github.com/nuejs/nue/actions/runs/11842921174?pr=397#summary-33002677721>

tipiirai:
It indeed catched broken links well. Is this safe to merge even though tit says "Some checks were not successful" ?

nobkd:
Yes. The checks aren't successful, because there are still broken links.

It should resolve itself over time, when fixing the links.

If there's a case where the link check fails, even if it shouldn't, I'll probably update the test soon after. On larger changes, there might be some of those cases, but I'll take care of it :)

nobkd:
I'll merge this now.
(And I might try to fix some more links later on)

tipiirai:
Thanks! I'll go trough the links soon

Issue #407, opened 2024-11-16
Add site.yaml schema to schemastore.org


JensRoland:
### Is your feature request or improvement related to a problem?
When editing the site.yaml file in VS Code, the editor mistakes it for an Ansible config file and marks every line as a syntax error.

### Solution you'd like
Adding the JSON Schema to [schemastore.org](https://www.schemastore.org/json/) and including a descriptor at the top of site.yaml should do the trick. Like so:

```yaml
# yaml-language-server: $schema=https://json.schemastore.org/nuejs-site.json

globals: ["@global"]
# ...
```

### Alternatives you've considered
No great alternatives.

### Additional context
I already did it and submitted it here: <https://github.com/SchemaStore/schemastore/pull/4218>

JensRoland:
Actually it would be better if Nue had a more specific alternative to `site.yaml` (maybe `nue.yaml`?) since schemastore.org understandably doesn't like to accept too-generic file match patterns.

Update: Added a PR for this: https://github.com/nuejs/nue/pull/408

tipiirai:
Hey. First, sorry — I've taken some freetime after the launch and haven't been too active here. 

My first though: damn Ansible! Second thought: isn't there a way around this in VS code? I code with Sublime, but when I open the site.yaml in VS code with default settings I get no such errors. 

Thirdly, I'm not a fan of "either / or" complexities in configuration and documentation so I really like to explore all the other options first. 

Thanks

JensRoland:
I completely understand.

As I see it, there are two good options that don't involve either/or:

1. Switch to using `nue.yaml` (or something else that is specific like that). This is a very small code change, and makes the file more clearly self-identifying, but it would be a breaking change, which sucks for Nue users.
2. Stick with `site.yaml` and leave a blank fileMatch in the schemastore.org submission. This still registers the schema, it just means editors can't auto-detect the schema based on filename. Then, by adding `# yaml-language-server: $schema=https://json.schemastore.org/nuejs-site.json` at the top of `site.yaml` you tell editors what schema to use.

joelschou:
I'm glad I decided to look through issues here before going any further in my investigation into how to do this. Having a schema for the config file is a huge benefit. Whether the config is `site.yaml` or `nue.yaml` is less of a concern (I prefer the latter, FWIW).

viktor-yakubiv:
Citing my https://github.com/nuejs/nue/pull/408#issuecomment-2501915669

[...]

PR #408, opened 2024-11-16
feat: add alternative nue.yaml config filename

JensRoland:
I have [submitted a JSON Schema](https://github.com/SchemaStore/schemastore/pull/4218) describing the Nue `site.yaml` file to schemastore.org in order to get editor support for the Nue config (autoexpand and validation). ~~However, since the `site.yaml` filename is generic, it is not ideal for file matching (needed for convenient auto-detection of the schema).~~

~~Thus, it would be nice to add an alternative filename, `nue.yaml`, which could be autodetected. I don't want to make a breaking change and propose changing it, but allowing both would work too.~~

~~This PR extends Nue to support either filename.~~

Update: The PR now simply adds a reference to the schema allowing editors to use it; and fixes a minor bug where `readData` would fail on missing path instead of returning.

Related: <https://github.com/nuejs/nue/issues/407>

### DIFF
--- a/packages/examples/simple-blog/site.yaml
+++ b/packages/examples/simple-blog/site.yaml
@@ -1,3 +1,4 @@
+# yaml-language-server: $schema=https://json.schemastore.org/nuejs-site.json
 
 globals: ["@global"]
 libs: ["@library"]
--- a/packages/nuekit/src/site.js
+++ b/packages/nuekit/src/site.js
@@ -41,6 +41,7 @@ export async function createSite(args) {
   }
 
   async function readData(path) {
+    if (!path) return
     try {
       const raw = await read(path)
       return yaml.load(raw)


viktor-yakubiv:
It would be great if even the site.yaml was also configurable. As an example, using the following command line option (`--config <path>`):

```bash
nue --config nue.yaml serve
```

JensRoland:
I have removed the multi-filename code as discussed in the issue: https://github.com/nuejs/nue/issues/407

This PR should be ready to merge

tipiirai:
wow! I like this. nice work. 

my only concern is how to keep the schema file up to date

tamaracha:
> my only concern is how to keep the schema file up to date

**Schema Design First** or **API Design First** approach :-)

PR #410, opened 2024-11-17
feat(action): site + example publish

nobkd:
This is still very much WIP, and completely untested.
I'll try working more on this, sometime next week.


nobkd:
Okay. The approach with pages won't work (because you can only define one domain for deploy).
But we can probably just compress our generated files push them over to the server via key based ssh and gh secrets and then unpack the files on the other side. (of course you could leave out the compression step, but it's probably much less space and time needed for transfer then.)

I'll rework this pr a bit and the we only need to add secrets containing the sensitive data.

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,91 @@
...
+  group: 'pages'
+  cancel-in-progress: false
+
+jobs:
+  prepare-ssh:
+    if: ${{ github.repository_owner == 'nuejs'}}
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Prepare SSH
```
JensRoland: Maybe simpler to use https://github.com/marketplace/actions/setup-ssh ?

nobkd: Thanks. (but I don't understand, why this action needs to use JS and extra libraries)

Edit: I'll look, if I use it, or just check if I can simplify the bash code a bit more instead

nobkd:
I might have to still add some code until the next release, so the example unpacking doesn't break.

Because I changed the unpack logic to not strip the first path segment. so with the change, the RC version would strip a path segment too much, cause the archives don't include the directory anymore.

Before: super directory `simple-blog` containing all the files is archived
With this PR: only contents of `simple-blog`, not the directory itself is archived

I will add that extra logic sometime next week (sorry that this is going forward so slowly), and remove it a while after the next release.

After that, this PR should be ready. It might just need some provider specific tweaks. Depends on how the files are saved on the web server.

I could also try changing the compression strategy back, but I think that might be harder, because I currently use git to only compress versioned files, and also don't include the directory itself.

I think I'm repreating myself, I should go do something different...

nobkd:
I'll do some last tests on my end this week, and might need to do some tweaks, so the files are uploaded to the correct host and unpacked to the right directory. I will probably need some provider info to do this. But until then I just do some basic tests running the action and triggering workflows in a private example.

nobkd:
Tests currently fail because of oven-sh/bun#16474

Edit: Converting to draft, until tests are fixed and current push to storage zone model is confirmed

.github/workflows/push-examples.yaml
```diff
@@ -0,0 +1,68 @@
...
+    if: ${{ github.repository_owner == 'nuejs' }}
+
+    strategy:
+      matrix:
+        dir: # dirs in "examples" to run build & push process for
+          - simple-blog
+
+    env:
+      dir: 'packages/examples/${{ matrix.dir }}/'
+      archive: 'site.tar.gz'
```
nobkd: Unused variable? Check and delete if applicable.

nobkd:
Note: I changed the archive name to `source.tar.gz` instead of using `<dirname>.tar.gz`. Currently the action generates both versions for compatibility.

I can also undo the change, to keep using the dir name variant. It doesn't really matter, but having the same file name for each example would be more expected I think.

Even after merging this, it is a change of one word (see `create.js`), so we can change it back before the next release, if we change our mind.

nobkd:
Todo: update action version for `ayeressian/bunnycdn-storage-deploy`.

PR #417, opened 2024-12-01
feat: `base` CLI option

### LINKED ISSUE #416: Add CLI option for `base`

nobkd:
fix #416

Add CLI option for `base`.

Currently not added to CLI help, because I think the following:

Maybe the logic for `base` has to be changed a bit, so you don't have to prefix it with a forward slash (`/`), but that's unrelated to this PR.
And currently you have to design your paths so, that they're relative. absolute paths are not prefixed properly in all the needed cases I think. (Also goes for custom `favicon` I think)


tipiirai:
I stopped using base personally so the lack of care is apparent.

PR #429, opened 2025-01-16
Update index.md

seneca:
A few changes of hyphens to en-dashes. The "On the final pahse" has a spelling mistake but also it doesn't read great. Maybe the author should rephrase it.


nobkd:
I agree, that rephrasing is a good idea.
Ping: @tipiirai

PR #430, opened 2025-01-17
Update data.md

seneca:
These points could be made either bold or the sentence after each point could go under them to make each point stand out better. Something to consider. I added a space after the first point because I didn't know how else to attach this note. The extra space should not be added, of course.


tipiirai:
Thanks. You can freely do formatting on these docs. I trust your instinct. 

This extra space is likely going to be removed by (my) text editor which auto-trims end of lines.

PR #446, opened 2025-01-18
Update index.md

seneca:
Perhaps, I don't understand what the author's intent here is, but in general a typo is a spelling mistake. I also added "to the server" as it seems something is needed here, but maybe this is not necessary.

### DIFF
--- a/packages/nuejs.org/vision/index.md
+++ b/packages/nuejs.org/vision/index.md
@@ -94,7 +94,7 @@ This is the final phase and it's time to squeeze everything out from the cloud. 
 nue push
 ```
 
-With this command a single typo that previously required a full rebuild now deploys in milliseconds through our global CDN. Content creators update sites directly through Markdown. Designers iterate on pages without build complexity. Marketing teams publish sophisticated landing pages instantly.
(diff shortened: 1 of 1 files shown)


nobkd:
Even your variant is still a bit unclear to me.
I mean, I get the gist of it, but the sentence still does not read well.
Maybe it should be rewritten completely?

seneca:
> Even your variant is still a bit unclear to me. I mean, I get the gist of it, but the sentence still does not read well. Maybe it should be rewritten completely?

seneca:
That would be ideal. I wasn't even sure what to suggest as an alternative. Maybe @tipiirai could have another go at this sentence.

tipiirai:
Here's another go for this:


The impact of `nue push` is dramatic: Even a one-character fix that would typically trigger a lengthy rebuild process now goes live within milliseconds through our global CDN. The system eliminates traditional build complexity, allowing content teams to update through clean Markdown, designers to refine pages directly, and marketing to launch sophisticated landing pages - all instantly and without technical overhead. For teams requiring version control, `nue push --git` provides the same blazing performance - deploying code changes in under 100ms compared to the typical 100-second builds of modern frameworks.

PR #447, opened 2025-01-19
Update content-tags.md

seneca:
Removed dashes from the explanation text. The text is already bulleted.


nobkd:
They're in an `[.options]` block too, so they get styled accordingly

PR #451, opened 2025-01-22
easier bun build error handling

nobkd:
also fixes new default `throw` behavior of bun v1.2


tipiirai:
good change. also happy that you merged this without a PR. no need when the improvement is as clear as it is here.

PR #460, opened 2025-01-26
feat(components): added support for all anchor attrs

alexpricedev:
> ⚠️ Sorry for all of the formatting changes. I have prettier installed and I checked it was picking up the `.prettierrc.yaml` but on save it still seemed to want to update a load of formatting....
Main changes are:
    - LOC:118-131 in `packages/nuekit/src/layout/components.js`
    - LOC: 68-98 in `CONTRIBUTING.md`

## Summary

This PR adds support for passing in all [allowed attrs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attributes) to an anchor tag (`<a>`). I also added some extra notes to the `CONTRIBUTING.md` _Local Development_ section as it took me a few minutes to figure out the bun linking and I thought the extra notes would help other first time contributors. 

The `a` attrs are useful for adding properties like `target="_blank"` to navigation items, eg:

```yaml
navigation:
  footer:
    - © Nue: https://nuejs.org/
      target: _blank
      rel: noopener noreferrer
```

| Rendered output |
|--------|
| image | 

## Notes

- [x] Test case added
- [x] Tests suite ran and completed with 100% pass rate

### DIFF
--- a/packages/nuekit/src/layout/components.js
+++ b/packages/nuekit/src/layout/components.js
@@ -1,4 +1,3 @@
-
 import { elem, parseSize, renderInline, renderIcon } from 'nuemark'
 
 import { readFileSync } from 'node:fs'
@@ -17,15 +16,16 @@ export function renderPageList(data) {
   return elem('ul', this.attr, pages.join('\n'))
 }
 
-
 // the "main" method called by the <navi/> tag
 export function renderNavi(data) {
   const { items, icon_dir } = data
 
-  return Array.isArray(items) ? renderNav({ items, icon_dir }) :
-    typeof items == 'object' ? renderMultiNav(items, data) : ''
+  return Array.isArray(items)
+    ? renderNav({ items, icon_dir })
+    : typeof items == 'object'
+    ? renderMultiNav(items, data)
+    : ''
 }
-
 
 function renderTOC(data) {
   const { document, attr } = data
@@ -38,7 +38,6 @@ function renderPrettyDate(date) {
   return elem('time', { datetime: date.toISOString() }, prettyDate(date))
 }
 
-
 // in Nue JS component format
 export function getLayoutComponents() {
   return [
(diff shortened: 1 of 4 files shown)


alexpricedev:
Not sure what I changed to break that test but I'll take a look!

alexpricedev:
Pretty sure nothing I changed should have effected that test suite but hopefully moving the `expect` outside of the `catch` will fix it 🤞

nobkd:
This isn't on you. Bun 1.2 changed the default behavior of `Bun.build` to `throw`, and we handled that differently before.

Please undo the test change :)
It should work now. (or when you rebase on master, not sure about that :shrug:)

packages/nuekit/src/layout/components.js
```diff
@@ -66,18 +65,22 @@ export function renderPage(page) {
...
-  const body = !thumb ? time + elem('a', { href: url }, h2 + p) :
-
-    // figure
-    elem('a', { href: url }, elem('figure',
-      elem('img', { src: thumb, loading: 'lazy' }) + elem('figcaption', time + h2 + p))
-  )
+  const body = !thumb
+    ? time + elem('a', { href: url }, h2 + p)
+    : // figure
+      elem(
```
tipiirai, line 71: I think we should stick to compact formatting in elem() calls consistently.

tipiirai:
The changes made here seem legit. In the future, I prefer more focused pull requests. For example, this PR should have been split in two: 1) updates to contributing 2) anchor attrs. More convenient. 

But in any case: thank you for the contribution! Will merge now

packages/nuekit/src/layout/components.js
alexpricedev: Agreed. I'll have a look at the prettier config and see if I can enforce that 👍

alexpricedev: I spent some time looking at this and really there isn't a good way of doing it without building a prettier plugin. I've never done that and would be up for trying it if you're interested?

PR #461, opened 2025-01-26
feat!: bun css support

nobkd:


nobkd:
We could also make the bun css option non-default first, and change that, if we find it stable. I have just played around with it, and it didn't show any issues, but maybe there are some.

I'm also still looking, if I can somehow replace the external array with an easier alternative, that marks css as internal and everything else as external...

tipiirai:
OH! I have completely missed this. Lets briefly talk before merging. So what exactly is Bun CSS support? Does this mean we can get rid of the heavy Lightning CSS support at some point? Now even? I would love to ditch LCSS because it makes 90% of all dependenices megabyte wise.

nobkd:
It's basically a rewrite of lightningcss in Zig by the Bun team, and it's included in Bun itself.

Yes, we could get rid of lightningcss (on bun).

Bun css is not as configurable, but I think the defaults set currently are probably alright.

Node should still work, if lcss gets installed locally. (like esbuild) 

I currently added a cli option, that switches to lightningcss. There's probably still a bit of documentation, that needs to get updated.

I still have some issues with the current implementation (e.g. the external array).
I'll try taking a stab at them later today.

tipiirai:
Hard to find any information about this. What happens with nested selectors? Are modern color functions supported? How the minifier works? etc. I really want this to replace Lightning CSS npm requirement. Would make Nue exceptionally slim. Developers could potentially enable LCSS support with local npm install.

nobkd:
As I said, Bun CSS is basically a rewrite of lightningcss:
https://github.com/oven-sh/bun/blob/93a89e5866fa85fea4026b084ec3b6efd1b6bdc0/src/css/README.md

I think this is the basic browser target, but I'm not sure:
https://github.com/oven-sh/bun/blob/93a89e5866fa85fea4026b084ec3b6efd1b6bdc0/src/css/targets.zig#L149-L154
To support those, nesting is currently unwrapped I guess.
I'm not sure, what happens with the modern color functions though.

I haven't seen any option to enable / disable any feature flags on `Bun.build` sadly. (even though it seems possible in the zig code)
https://github.com/oven-sh/bun/blob/93a89e5866fa85fea4026b084ec3b6efd1b6bdc0/src/css/targets.zig#L12-L18
https://github.com/oven-sh/bun/blob/93a89e5866fa85fea4026b084ec3b6efd1b6bdc0/src/css/targets.zig#L116-L136


Note: I dind't get the css bundler to properly work with versions of Bun <1.2, so I changed the engines. I hope to still get it to work, but I can't really promise anything.

tipiirai:
Okay, wow. Looking at the Bun code they seem to be taking CSS seriously. A major Zig engineering project. Happy to merge this and Bun 1.2 requirement is not an issue. I would even depreciate Node for this — not a fan of branching to ESBuild and Lightning CSS just to get Node working. Bun is such a stable product these days.  Nue website CSS is a good test framework for this feature: if all looks good, then we're good to go. Thoughts?

nobkd:
I didn't have any issues with the site's CSS, so I'd say this is ready.
Would keep the possibility for node users for now, even though it's not pretty.

Some css options don't make sense with Bun, as Bun.build doesn't expose e.g. feature flags like native nesting (yet)... Added this to docs.

tipiirai:
Exciting feature. Thank you for the hard work.

PR #465, opened 2025-01-27
fix(glow): handle inline /* ... */ comment

### LINKED ISSUE #463: Glow truncates code; output incomplete
### Describe the Bug

Glow truncates code; output is incomplete.

```
import { glow } from 'nue-glow';

const code = `const machine = defineMachine<State, Event>({
  states: {
    [stateName]: {
      on: {
        [EVENT_NAME]: { 
          to: 'nextStateName',
          /* ... other options ... */ 
        },
        // ...
      },
    },
    // ...
  },
});
`;

const html = glow(code, { language: 'typescript', numbered: true });

console.info(html);

const appDiv: HTMLElement = document.getElementById('app');
appDiv.innerHTML = `<pre>${html}</pre>`;
```

Rendered output is 

```
const machine = defineMachine<State, Event>({
  states: {
    [stateName]: {
      on: {
        [EVENT_NAME]: { 
          to: 'nextStateName',
```

Live demo: https://stackblitz.com/edit/typescript-playground-tenktfl9?file=index.ts

### Environment

Web, node

"nue-glow": "^0.2.2"

### Minimal Reproduction

As above

maurice:
Fixes #463 

Please shout if there is anything you are unhappy with.

### DIFF
--- a/packages/glow/src/glow.js
+++ b/packages/glow/src/glow.js
@@ -39,6 +39,9 @@ const HTML_TAGS = [
 
   // line comment
   { tag: 'sup', re: /# .+/ },
+  
+  // inline comment
+  { tag: 'sup', re: /\/\*.*?\*\//g },
 
   { tag: 'label', re: /\[([a-z\-]+)/g, lang: ['md', 'toml'], shift: true },
 
@@ -223,7 +226,7 @@ export function renderRow(row, lang, mark=true) {
 
 
 // comment start & end
-const COMMENT = [/(\/\*|^ *{# |<!--|'''|=begin)/, /(\*\/|#}|-->|'''|=end)$/]
+const COMMENT = [/(\/\*|^ *{# |<!--|'''|=begin)/, /(\*\/|#}|-->|'''|=end)/]
 
 export function parseSyntax(lines, lang, prefix = true) {
   const [comm_start, comm_end] = COMMENT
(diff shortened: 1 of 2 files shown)


nobkd:
Great work so far!

I added `if (comment) endComment()` to line 270 in `glow.js`, so the comment gets closed, at the end of the lines even if there is no closing end (or no "block" closing end, to be specific).

In the following example, even though no code is truncated (with the above addition), the end is still a comment, because the closing was "inline" vs the opening was "block":

Try:
```js
/*
hello
*/ world
```

Do you have a simple fix idea?

(Btw. there are still the other multiline comment types, (e.g. `<!-- hello --> world`) that also need an inline variant.

I guess, this grows a bit in complexity... 

I'm not sure, how to solve this, but you can try giving it a stab, if you'd like :)

maurice:
> Great work so far!

Thanks for the encouragement 😊

> In the following example, even though no code is truncated (with the above addition), the end is still a comment, because the closing was "inline" vs the opening was "block":

Good test case 👍 

> (Btw. there are still the other multiline comment types, (e.g. `<!-- hello --> world`) that also need an inline variant.

Yeah that's true, and we should be able to use the same technique for all really.

> I'm not sure, how to solve this, but you can try giving it a stab, if you'd like :)

I've updated the PR to handle your example (but not the other multiline comment types). I've tried to be sympathetic to the original implementation, but given the assumptions that were made originally (specifically, it's a line-by-line transformation, and each line is assumed to be either a comment or not), it's a bit hacky.

I'm not expecting this to be merged as-is, but we could talk about where to go from here.

I'm a bit concerned about making more significant changes because of the ultra minimal test suite, and there are definitely some lines I can comment-out and no tests fail 😟

packages/glow/src/glow.js
```diff
@@ -232,16 +235,20 @@ export function parseSyntax(lines, lang, prefix = true) {
...
   }
 
-  lines.forEach((line, i) => {
+  for (let i = 0; i < lines.length; i++) {
+    let line = lines[i]
     if (!comment) {
-      if (comm_start.test(line)) {
+      const start_comm = comm_start.exec(line)
+      const end_comm = comm_end.exec(line)
+      if (start_comm && start_comm.index === 0 && (!end_comm || line.length === end_comm.index + end_comm[0].length)) {
```
tipiirai, line 248: @maurice : please note that Nue codebase prefers loose equality:

https://github.com/nuejs/nue/blob/master/CONTRIBUTING.md#code-style

maurice: Yep I did read that and was really fighting with my habits to not add semi-colons, manually format code, and generally keep it consistent, but obviously missed this.

TBH this change was meant to stimulate some discussion, but it seems you guys have other priorities right now, which is fine.

In the meantime I switched to [shiki](https://shiki.style/) for syntax highlighting, so I'm going to close this PR now

PR #466, opened 2025-01-27
feat: added a blank nue example project for use with 'create' cli

alexpricedev:
## Summary

This PR adds a new _"blank"_ example nue project to the `create` CLI as after starting two `simple-blog` projects and having to delete most of the dirs I wanted a cleaner starting place.

### Screenshots

Screenshot 2025-01-27 at 20 22 56

## Changes Made 

1. Adds a new `blank` project in `packages/examples/blank/`
2. Adds `blank` to the `push-examples` GHA
3. Adds support for `blank` in the create CLI js file

#### Notes

1. I've not actually be able to test that the CLI part works as the bundle needs to be pushed up first, but I assume all that pipeline works as expected 🤞 
2. Whilst this "blank" template could be _emptier_ I wanted to retain the important file structure elements and _some_ example code as a jumping off point for new and seasoned users


nobkd:
The action code etc. looks fine.

You may want to either add a welcome page or change line 13 in create.js to open `/` instead of `/welcome/` for the blank example. Not completely sure how to handle that, but maybe you have an suggestion. (idea: const object variable that lists project starter names as keys and opened page as value. Is used to check if template exists, and which page to start on. I'm open to more ideas) 

In the end, it's up to @tipiirai to decide to add this example (and secrets and storage / pull zone)

nobkd (approved):

alexpricedev:
Great shout about `/welcome`. I'll have a look at that!

alexpricedev:
I've added a redirect from `/welcome` to `/`.

Doesn't feel like a super clean solution but it was good for me to play with some of the scripting stuff and might be useful for others?

Also up for creating an _opening page mapper_ if thats a better solution. Not sure if we'll be able to tell which template was used though. Maybe on first load? 🤔

tipiirai:
I'll check this out after the next major release about single-page applications, which will add a prime template to the repository. We need to strategize and document what templates are available and why.

alexpricedev:
Closing this out to reduce PR noise

Issue #379, opened 2024-10-03
Nuemark2.0 issues


tomByrer:
Doing a quick code review of your dev branch, I am confused if the Blocks or Inline is scanned first?  If Blocks, then it seems this code:
```markdown
 *** Bold Italic ***
 ```
 Would render as an `<hr>` since you are just scanning first few characters.?
 
 Also be aware that [CommonMark is VERY flexible for `<hr>`](https://spec.commonmark.org/0.31.2/#thematic-breaks).  I personally don't think you need to cover every edge case, but it could help to document "Only the common Markdown/CommonMark rules are covered."

nobkd:
@tipiirai 

`[![](someimagelink)](somelink)` in nuemark2 breaks the nue server and generator.

I used `[![](/img/blog-build.png)](/)` on a random page in the `docs`, and the server just hangs indefinitely

I know, that this is possible using markdown extensions (e.g. `[! /img/blog-build.png href="/"]`), but the default Markdown variant should still work imo.

---

The next ones generate and don't let the server hang, but still generate unexpected output:

| Markdown | HTML |
|---|---|
| `[[! yo.svg]](/)` | `<p>&lt;[! custom=&quot;[!&quot;&gt;<script type="application/json">{"_":"yo.svg"}</script>](/)</p>` |
| `[[my-tag]](/)` | `<p>&lt;[my-tag custom=&quot;[my-tag&quot;&gt;](/)</p>` |

nobkd:
Btw I tried your example and this ` *** Bold Italic ***` renders to nothing (mabye a section split??? idk) in my test on a doc page.

I also tested `*** Bold Italic ***` and it generates `<p><strong>* Bold Italic</strong>*</p>` but I think it should generate `<p>*** Bold Italic ***</p>`.

nobkd:
@tipiirai just so you know, dev branch fails to build the docs (when there's no current `.dist`) since [`table`](https://github.com/nuejs/nue/commit/aac80149b3a6b118304708f5b632b5c65ea42dfe) changes:

Files where it fails on `[table]`:
- `packages/nuejs.org/docs/syntax-highlighting.md`
- `packages/nuejs.org/blog/rethinking-reactivity/index.md`

Error message:
```js
227 | 
228 |   const html = rows.map((row, i) => {
229 |     const is_head = head && i == 0
230 |     const is_foot = table.foot && i > 1 && i == rows.length - 1
231 | 
[...]

PR #477, opened 2025-02-05
feat(nuemark)!: proposal to use different section separator

nobkd:
- link https://github.com/nuejs/nue/issues/379#issuecomment-2528041394

Increase cmark compatibility by a tiny bit.

Use `+++` as section separator, makes it clear, that no traditional `<hr>` is used

```rst
hello section 1

+++

hello section 2
```

### DIFF
--- a/packages/examples/simple-blog/contact/index.md
+++ b/packages/examples/simple-blog/contact/index.md
@@ -10,7 +10,7 @@ title: Contact me
   
   [contact-me]
 
-  ---
+  +++
   [image]
     src: /img/profile.jpg
     size: 460 x 460
--- a/packages/nuejs.org/blog/rethinking-reactivity/index.md
+++ b/packages/nuejs.org/blog/rethinking-reactivity/index.md
@@ -82,7 +82,7 @@ To understand this choice we must go back in time. The world used to be slightly
   - Accessibility
   - UI libraries
 
-  ---
+  +++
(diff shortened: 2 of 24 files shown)


tipiirai:
Damn. A tough decision. I understand where you're coming from. I think feature changes of this caliber should be discussed first. I feel +++ puts this closer to hackers and not content writers, which is the goal. I also don't value commonmark very highly. Content development experience is the priority. Do you think +++ is better than === ?

nobkd:
I thought about **new** (section) on the *plus* symbol, but it doesn't really matter. We can also use `===` as section splitter.
But I wouldn't use  `---`, because I always used it as `<hr>` (and didn't know about e.g. `***` as valid alternative) and it feels most intuitive for `<hr>` (for me) too.

I can see both `+++` and `===` as viable/valid options. The decision is, what you like more, though. (or allow both 🤷‍♂️ :)
I'll change the code depending on your preference.

Alternatively I can also close this one and we can discuss this first somewhere else.

tipiirai:
My original idea with --- was to use the same separator as in front matter. One syntax to remember. But I feel this is often used as `<hr>`. This is a compromise, but I'm leaning towards "---" as section separator. It feels so natural, especially inside blocks.

nobkd:
might come back to this, but this is by no means priority in my mind

PR #478, opened 2025-02-06
docs: fix some links

nobkd:
> [!Note]
> I also changed some links which pointed to `content-syntax`, since you moved it to `_content-syntax.md`.
> Not sure, if this was accidental, or if you want to remove that file in the future, though!

Also not sure, what to do about most of the broken links that are left over. Since the create repo doesn't exist anymore, the links to the examples/templates are, and will obviously stay broken.

Please check, if you think the changed links are pointing to places you intended them to point to.


nobkd:
Might reopen for once dev is done. But maybe I will do fixes on dev one after one, or so.

PR #479, opened 2025-02-06
feat: use native `import.meta.resolve`

nobkd:
- requires node v20.8+
- gets rid of an extra dependency

Edit: not sure, why this fails in CI. Closing, until I find out.


nobkd:
Is a jest issue :(
See https://github.com/jestjs/jest/issues/15330

PR #480, opened 2025-02-06
feat: enable colored help text on windows, support `NO_COLOR` env var

nobkd:
- enable colored help text on windows
- support [`NO_COLOR`](https://no-color.org/) env variable

### DIFF
--- a/packages/nuekit/src/cli-help.js
+++ b/packages/nuekit/src/cli-help.js
@@ -64,7 +64,7 @@ function formatLine(line) {
 }
 
 export function getHelp() {
-  return !(process.env.TERM || process.platform == 'win32') ? HELP : HELP.split('\n').map(line => {
+  return HELP.split('\n').map(line => {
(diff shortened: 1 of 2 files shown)


packages/nuekit/src/cli-help.js
```diff
@@ -64,7 +64,7 @@ function formatLine(line) {
 }
 
 export function getHelp() {
-  return !process.env.TERM ? HELP : HELP.split('\n').map(line => {
+  return !(process.env.TERM || process.platform == 'win32') ? HELP : HELP.split('\n').map(line => {
```
nobkd:
```suggestion
  return process.env.NO_COLOR || !(process.env.TERM || process.platform == 'win32') ? HELP : HELP.split('\n').map(line => {
```

nobkd:
Maybe move all the color handling to the coloring functions instead of handling it separately in cli-help

PR #485, opened 2025-02-11
Collection of small create, cli, version, helper changes

nobkd:
- set & show nue version before `create`
- simpler usage of `version` by running function on startup
- prepare `create` for more templates
- pass more args to `create` (e.g. usage of `nue create simple-blog -p` works now.)
- no more manual version file entries: `@nue/.v<current nuekit version>` file is now automatically created
- add `docs` command for quick access (can remove if unwanted)
- add `--open` option, to allow opening website on nue serve
- ...

If this gets accepted, #466 might need some changes

### DIFF
--- a/packages/nuekit/src/cli-help.js
+++ b/packages/nuekit/src/cli-help.js
@@ -1,4 +1,4 @@
-import { colors, openUrl, version } from './util.js'
+import { colors, version } from './util.js'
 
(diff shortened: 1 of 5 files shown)


tipiirai:
@nobkd looks solid. Feel free to merge after resolving the conflict. Thanks

PR #488, opened 2025-02-18
Drop `import-meta-resolve` & `es-main`

nobkd:
This is currently WIP. Tests probably fail on `jest` too.

This will change the behavior of importing `esbuild` and `lightningcss` and probably other imported modules slightly, because they now have to be installed relative to the `nuekit` installation, and not in the nue project. If someone wants to have modules only for a specific project now, they'll also need the `nuekit` installation in the project itself. This is just a hunch, I'd have to check that more.

TODO:
- [x] drop `es-main`
- [x] drop `import-meta-resolve`
- [x] ignore jest peer dependency conflict on npm
- [x] check if resolve behavior is really different now for project js imports

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -43,6 +43,8 @@ jobs:
 
       # Testing `node+jest`
       - uses: actions/setup-node@v4
+        with:
+          node-version: '>= 20.8'
         if: ${{ matrix.tool == 'node+jest' }}
 
       - name: Install and test with Node
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -50,6 +50,6 @@ jobs:
       - name: Install and test with Node
         run: |
           node -v && npm -v
-          npm install --no-save jest jest-extended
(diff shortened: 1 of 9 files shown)


tipiirai:
Makes sense. I think we need to have the local installation option available anyway. Not everyone likes the global option.

nobkd:
Confirmed, that packages (at least esbuild, so I think this goes for all) now need to be next to nuekit.

This difference could probably only be solved by using something like `createRequire` from `node:module` or so.

nobkd:
I somehow got over this, and tested only a few things, will finally take a closer look on Wednesday, after I'm done with my next exam.

nobkd:
In conclusion, I think, you can install nuekit globally, and if you want to use node, have to install lightningcss & esbuild next to it.
But you can still have the other project's packages only in the project modules (not next to global nue). Project module resolving seems to work fine.
I think that's because esbuild/lcss are used from within nuekit and project deps are only used in the project itself and not from nue

nobkd:
@tipiirai if you agree, I'll merge this.

PR #489, opened 2025-02-19
feat: gh create

nobkd:
basic github create.

- needs more checks & testing
- won't start project automatically, could be a security issue

TODO:
- [x] check if `--strip-components` needs to be removed for gh (maybe change example .tar.gz build accordingly)
  Update: has files in a sub-dir (`<username>-<reponame>-<commithash>`) in the archive, so the stripping is correct :)

### DIFF
--- a/packages/nuekit/src/create.js
+++ b/packages/nuekit/src/create.js
@@ -20,7 +20,7 @@ async function serve(args) {
 }
 
 export async function create(args = {}) {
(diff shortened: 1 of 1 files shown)


tipiirai:
True. Starting the project automatically is probably not a good idea due to security concerns.

tipiirai:
wow. this came out beautifully simple!

PR #491, opened 2025-02-19
fix: better `argv` stripping, not redefining `cmd` anymore

nobkd:
- fixes e.g. `nue create simple-blog -r path/nue` (which ends on `/nue` and therefore was filtered), there shouldn't be more than two argv's needed to be stripped, i think. (will check it again though), maybe we can just strip the first two argv always?
- fixes e.g. `nue create simple-blog -r init stuff` (which previously redefined the command to `init`)

### DIFF
--- a/packages/nuekit/src/cli.js
+++ b/packages/nuekit/src/cli.js
@@ -33,7 +33,7 @@ export function getArgs(argv) {
       args.test = true
 
       // command
-    } else if (commands.includes(arg)) {
+    } else if (!args.cmd && commands.includes(arg)) {
       args.cmd = arg
 
       // options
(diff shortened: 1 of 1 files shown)


nobkd:
I'm sure, that it's always `js-runner cli.js-file real --arguments`, but test it yourself too.

I even tried passing arguments to or using a subcommand of the js runner, and they were not in argv e.g. `node --experimental-vm-modules $(which nue) ...` -> `['node', 'path/to/cli.js or so', '...']`

nobkd:
I can also move the `slice(2)` part back to `getArgs`, if preferred.

PR #494, opened 2025-02-22
fix: added special case in proxy element getter for attrs

alexpricedev:
## Summary 👀

This PR fixes an issue with `nue.js` that I found whilst playing with the demos. It doesn't include any tests though, I'm working on that separately. If this should wait for more automated tests I totally understand 🙏 

### Screenshots 📸

| Slot demo without fix | Slot demo with fix |
|--------|--------|
| image | image | 

## Changes Made 💪

1. Added a special case in the proxy component for `$attrs`

### Testing instructions 🧪

1. Compile and run the `packages/nuejs/test/client/basics.dhtml` and check the `slot-test` component and it's nested field.

#### Notes 📋

There might be a better solution for this. For example, I'm not sure if this would work for a component which was one level deeper in the tree. I suspect there is some refinement we can do with the recursive calls to `createApp` so the context is shared for these highly nested components. I'll have to learn more about the code before I do that though!


tipiirai:
Awesome! This is good example case why we need tests, because a DOM element can have a value property and a value attribute and both needs to be handled properly.

PR #499, opened 2025-03-05
RFC: nue.js domino based tests

alexpricedev:
## Overview

This PR introduces a new testing infrastructure for Nue.js components using [domino](https://github.com/fgnass/domino) to simulate a browser environment. This allows us to run interactive component tests automatically in a lightweight and controlled DOM environment.

🙋‍♂️ It's a POC and I welcome comments on both the approach and structure of the test suite.

### Goals

1. Automated, interactive DOM tests
2. Retain the ability to do manual browser testing, and ideally use the same components/files as the automated tests

### New Test Suites

Added initial test suites for some basic & core features:

1. Conditional Rendering (`test-ifs/`)
3. Click Events (`test-clicks/`)
4. Component References (`test-refs/`)

### Running the tests

You can run the tests from the `nuejs` dir with `bun test` or if you run `bun compile` you can view them in the browser (assuming you're running a local server of some kind).

## Nue.js Changes

I tried to keep changes to the core lib to a minimum for now, but I did have to replace `new Text('')` with `document.createTextNode('')` for better DOM compatibility.

## Dependencies

Added:
```
"devDependencies": {
  "domino": "^2.1.6"
}
```

## Next Steps
- Add more test coverage for remaining and more complex features
- Remove old manual browser test files
- Add this testing into CI


nobkd:
Here are some changes I made, (without really looking into the tests themselves, though. Will take a closer look, probably on Wednesday): alexpricedev/nue#1

tipiirai:
Thanks @alexpricedev — stellar work. Lets discuss this week

tipiirai:
Perhaps domino should be a peerDependency? I think that makes sense, since most installs are for users and not Nue developers

nobkd:
- Peer dependencies are for users. They can be marked optional, so the users have to install them themselves. If they aren't marked optional, the users must already have installed them, or the package manager complains, I think. 
- Dev dependencies are for developers. They are not installed, when installing normally from npm, afaik
- There are also optional dependencies, but they get automatically installed, if they are available for the platform.

So domino should stay a dev dependency, and esbuild and lcss are correct as both peer and dev.

It might be, that I got something wrong in my list above though.

PR #500, opened 2025-03-06
Mention optional / test deps

nobkd:


tipiirai:
peerDependencies seems like the perfect system for ESBuild and Lightning CSS

PR #504, opened 2025-03-11
Domino test fixes

nobkd:
Make tests work on jest, esm compatible dirname, etc. Same as alexpricedev/nue#1

Copied text:

---

- automatic `bun:test` imports <= fixes jest
- unwrap used imports
- use esm compatibility dirname <= I think, also fixes jest

(
just for notice: if you set the `test` script, and run `bun test`, the test script is not actually used. You'd need to use `bun run test` / `npm test` (which we currently use to have an easier `jest` testing command)
:)

If you want to have the `compile` script running on tests or the `timeout` added back, I can do so. (using `bunfig.toml` and `jest`'s config)

I'll probably take a closer look at your tests on Wednesday. First wanted to have the tests run on both bun and jest, and have a similar style to the other test files. I hope that's fine with you.

If there's something you'd like different in my changes, let me know!


nobkd:
merging.

PR #507, opened 2025-03-12
test(nuejs-core): `for`-tests

nobkd:
@alexpricedev maybe you can also take a look, if this looks fine to you. I roughly oriented myself on your tests.

I can also split *array funcs* to *n* tests, so only one function is tested at a time. Would probably be more deterministic / ensure their functionality. 

---

PS: what do you think about a template `index.html` for the manual browser tests and creating the needed `index.html`s with `compile` script to `dist`, next to the compiled components?
EDIT: added this

---

Another edit: I think, the for loops are currently not re-run, if we change content of a nested item. I'm gonna take a look at that tomorrow.
For now making this WIP.

Update: I don't think there's any easy solution to this, so not going to follow this here anymore. maybe in the future.


alexpricedev:
I need to sort my notifications out as I only just saw this when looking manually! Maybe you can DM me on Slack if you need my 👀 ?

Will take a look tomorrow 🤙

nobkd:
Oh, no worries. I changed this to wip, because I still wanted to check some stuff.
But I can message you on slack as well, when I mention you sometime in the future :)

alexpricedev (commented):
Super cool, nice work :) I'd approve it in its current state for sure.

nobkd:
Nice. I changed my mind about adding more to this PR, and will merge now, since I made no significant changes after the review.

PR #508, opened 2025-03-14
feat: basic ignore support

### LINKED ISSUE #198: Allow users to pass additional IGNORE dirs and filenames from commandline

nobkd:
- fix #198 
- fix #503
- ref #202

This is a very basic implementation. I'll probably do some changes on this.
Would need to be added to the yaml config mentions (docs / schemastore)

### DIFF
--- a/packages/nuejs.org/docs/command-line-interface.md
+++ b/packages/nuejs.org/docs/command-line-interface.md
@@ -40,11 +40,15 @@ Usage
 
 - **-n or --dry-run**: Displays what would be built without actually creating any outputs. This is useful for previewing changes.
 
-- **-b or --esbuild**: Uses esbuild as the bundler for your assets. Note that you need to install esbuild manually for this to work.
+- **-b or --esbuild**: Uses `esbuild` as the bundler for your assets. Note that you need to install `esbuild` manually for this to work.
 
-- **-l or --lcss**: Uses lightningcss as the minifier for your CSS. Note that you need to install lightningcss manually for this to work.
+- **-l or --lcss**: Uses `lightningcss` as the minifier for your CSS. Note that you need to install `lightningcss` manually for this to work.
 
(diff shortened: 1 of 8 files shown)


nobkd:
Solved in nuekit 2.0 beta using
```yaml
site:
  skip: [...]
```

PR #514, opened 2025-03-20
docs: "skip to content" button

nobkd:
On first <kbd>tab</kbd> you get a link to "jump to the main content":

before first tab:
grafik
after first tab:
grafik

Prevents having to tab through all navigation elements (especially on the docs pages)

Best to just try it out locally.

### DIFF
--- a/packages/nuejs.org/@global/layout.html
+++ b/packages/nuejs.org/@global/layout.html
@@ -1,6 +1,7 @@
 
 <!-- site header -->
 <header>
+  <a class="button skip" href="#main-content">Jump to main content</a>
   <navi :items="globalnav.main"/>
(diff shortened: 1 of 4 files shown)


packages/nuejs.org/@global/layout.html
```diff
@@ -1,6 +1,7 @@
 
 <!-- site header -->
 <header>
+  <a class="button skip" href="#main-content">Jump to main content</a>
```
nobkd:
```suggestion
  <a class="button skip" href="#main-content">Skip to content</a>
```

nobkd:
I might reopen this later on. after updating to the new nue

PR #520, opened 2025-03-26
feat(create): fetch available templates list from gh api

nobkd:


nobkd:
Might try again in the future.

PR #525, opened 2025-04-01
fix: make simple-mpa javascript example not stack overflow

akx:


tipiirai:
Nice. Did you manage test this locally?

akx:
> Nice. Did you manage test this locally?

Yes – I applied the same change locally into the simple-mpa example and replaced the mock chunk files with the big chunks, and it works fine.

PR #530, opened 2025-04-01
Update single-page-apps.md

seneca:
This refers to "an application" in singular so it should be "pushes" or pluralise applications and then push can stay as is?


nobkd:
Thanks!

PR #531, opened 2025-04-01
Update single-page-apps.md

seneca:
Is there a reason to opt for controllers to be plural and model and view singular?


nobkd:
You mean here https://github.com/nuejs/nue/blob/6254990a50775d841fe2ba7cbc37396175ad5c4e/packages/nuejs.org/docs/single-page-apps.md?plain=1#L30 and https://github.com/nuejs/nue/blob/6254990a50775d841fe2ba7cbc37396175ad5c4e/packages/nuejs.org/docs/single-page-apps.md?plain=1#L41 here?

seneca:
Yes, just curious why other folders like "model" or "view" is not pluralised as well.

tipiirai:
There is. A SPA can have multiple contorllers. Model is always singular. View is also thought out as a single view, consisting of multiple components.

nobkd:
Closing, as the question seems to be resolved.

PR #534, opened 2025-04-01
Update content-focused-apps.md

seneca:
Another paragraph that looks like a comment and is out of place?


tipiirai:
Thanks!

nobkd:
@tipiirai I think @seneca wanted you to check if the paragraph was out of place in your opinion too?
There was no real change on the content up till now, just a marker to have the content that was meant, be highlighted.

Same for: #533 (and #531)

PR #546, opened 2025-04-06
fix: nuekit create script call without args

lucas-prado-silva:
When calling `nue create` without any parameters, the current behavior is the following: 

```
nue create

✓ Nue 1.0.0-RC.3 • Bun 1.2.8
22 |   if (!args.debug) openUrl(`http://localhost:${nue.port}/${tmpl.open || ''}`)
23 |   return terminate
24 | }
25 |
26 | export async function create(args = {}) {
27 |   if (!args.name) args.name = args.paths.shift().split('/').filter(Boolean).join('/') || 'simple-blog'
                                                    ^
TypeError: undefined is not an object (evaluating 'args.paths.shift().split')
      at <anonymous> (/Users/lucas-prado-silva/.bun/install/global/node_modules/nuekit/src/create.js:27:48)
      at create (/Users/lucas-prado-silva/.bun/install/global/node_modules/nuekit/src/create.js:26:30)
      at <anonymous> (/Users/lucas-prado-silva/.bun/install/global/node_modules/nuekit/src/cli.js:101:18)
```

This PR makes sure that the `simple-blog` template is used as the default when no arguments are used, which seems to be what the previous code wanted to achieve.

### DIFF
--- a/packages/nuekit/src/create.js
+++ b/packages/nuekit/src/create.js
@@ -24,7 +24,7 @@ async function serve(args) {
 }
 
 export async function create(args = {}) {
-  if (!args.name) args.name = args.paths.shift().split('/').filter(Boolean).join('/') || 'simple-blog'
+  if (!args.name) args.name = args.paths.shift()?.split('/').filter(Boolean).join('/') || 'simple-blog'
   if (!args.root) args.root = args.paths.shift() || args.name.replaceAll('/', '-')
 
   // debug mode with: `nue create test`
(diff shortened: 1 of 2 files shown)


packages/nuekit/src/create.js
```diff
@@ -24,7 +24,10 @@ async function serve(args) {
 }
 
 export async function create(args = {}) {
-  if (!args.name) args.name = args.paths.shift().split('/').filter(Boolean).join('/') || 'simple-blog'
+  if (!args.name) {
+    const first_path = args.paths.shift()
+    args.name = first_path ? first_path.split('/').filter(Boolean).join('/') : 'simple-blog'
+  }
```
nobkd:
```suggestion
  if (!args.name) args.name = args.paths.shift()?.split('/')?.filter(Boolean)?.join('/') || 'simple-blog'
```

What do you think about this? Or is it not readable enough?

lucas-prado-silva: looks good to me 👍

nobkd:
Thank you!

PR #549, opened 2025-04-07
docs: accessibility improvements

nobkd:


packages/nuejs.org/@lib/video.css
```diff
@@ -48,7 +47,7 @@
   height: 4.2em;
   width: 4.2em;
   left: 50%;
-  top: 45%;
+  top: calc(50% - 2.1em);
```
nobkd: Todo: Use variable e.g. `--size` for width, height and top: `calc(50% - var(--size) / 2)`. Check if left: 50% really centers correctly

PR #556, opened 2025-04-19
site: blog feed

### LINKED ISSUE #283: RSS Feed Generation
Add option to generate feed(s) RSS/Atom/… for specified directories.

Helpful for e.g. blogs / news pages.

Ref: nuejs/www#53

nobkd:
- partially fixes #283

Todo
- [ ] use Atom format

### DIFF
--- a/packages/nuekit/src/layout/components.js
+++ b/packages/nuekit/src/layout/components.js
@@ -13,32 +13,38 @@ export function toFeed(data) {
     elem('guid', `${data.origin}${e.url}`),
     elem('link', `${data.origin}${e.url}`),
     elem('pubDate', e.date.toUTCString()),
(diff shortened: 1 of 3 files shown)


tipiirai:
I think the feed should be created when the site is built and with a --rss  (or --feed) configuration option. Thoughts?

nobkd:
Just playing around for now... First wanted to generate a valid feed, then looking further, so nothing thought-out yet

cevdetta:
I think it's good to support just one format (RSS 2.0 vs Atom 1.0) which should be Atom;
- to avoid confusions,
- Atom is way [better](https://www.intertwingly.net/wiki/pie/Rss20AndAtom10Compared),
- development of it would be more straightforward.

PR #574, opened 2025-05-07
Expand MIME type map in web server to broaden asset type support with fallback (fonts, source maps, etc)

albedosehen:
Updates the web server with additional MIME types in `TYPES` and a fallback for unknown file extensions. Previously, requesting files with unrecognized extensions (e.g. .map) caused a TypeError: Invalid value "undefined" for header "content-type" to be thrown when serving a file. This results in incomplete responses from the web server.

Adds the following changes:
- `.map`, `.woff2`, `.webmanifest`, `.wasm`, fonts, media, etc.
- Updated `jpg` and `jpeg` to use `image/jpeg` per [rfc3745](https://www.rfc-editor.org/rfc/rfc3745)
- Fallback to prevent/resolve header errors for undefined `content-type` using `application/octet-stream` as per [rfc2046](https://datatracker.ietf.org/doc/html/rfc2046#section-4.5.1)
- Test to ensure all known MIME types are recognized and that unknown file extensions return `application/octet-stream`


packages/nuekit/test/nueserver.test.js
```diff
@@ -0,0 +1,45 @@
+import { writeFile, rm, mkdir } from 'node:fs/promises'
+import { join } from 'node:path'
+import { createServer, TYPES } from '../src/nueserver.js'
+
+const TEST_DIR = join(import.meta.dir, 'tmp-mime-types')
```
nobkd: Our lowest supported node version doesn't support `import.meta.dir` yet.

Please use something like the following:
`const srcdir = dirname(fileURLToPath(import.meta.url))`

I'll take a look at the added test later today.

nobkd (changes requested):

packages/nuekit/test/nueserver.test.js
albedosehen: Updated to use `fileURLToPath(import.meta.url)` as suggested.

nobkd:
Sorry about the delay, I'll come back to this, as I still want to do a manual check on this change.

nobkd (approved):

nobkd:
Thanks for the contribution. I was a bit preoccupied with other stuff the last days, so I just came back around to the last checks now.

PR #582, opened 2025-06-12
fix typo

neeko-cat:
I fixed an oopsie


nobkd:
Thanks for fixing this

Issue #599, opened 2025-08-01
Build directories are not fully wiped on new build


NickSdot:
### Describe the Bug
Build directories are not fully wiped on new builds. This can lead to leftover files in build. Imagine this:

- Build A has a post `typo.html`
- You realise the typo, and rename the file to `fixed.html`
- Build B will have `typo.html` and `fixed.html`

### Solution you'd like
Build subdirectories should be wiped on new builds.

### Environment
- MacOS latest
- Nuekit version & JS runtime: Nue 1.0.0-RC.4 • Bun 1.2.19

Issue #598, opened 2025-07-31
Collection urls should not be ambigious


NickSdot:
### Is your feature request or improvement related to a problem?
Yes. A collection 'blog' can be accessed with and without trailing slash:

```txt
https://example.com/blog/
https://example.com/blog
```

Strictly speaking, this is duplicate content for search engines. There are three potential solutions:

1. Decide for one of the above interally; 404 on the other.
2. Treat it the same way was posts and append `.html`: https://example.com/blog.html
3. Define `canonical` (IMO, avoidable overhead.

### Solution you'd like
Personally, I cannot come up with a reason why post and collection pages should be treated differently. Given that, and assuming that it is undesired to drop `.html` from actual posts, my prefered solution would be to append `.html` to non-posts.

### Alternatives you've considered
This surely can be solved by server-side rewrite rules. However, I believe Nue should handle this cleanly out of the box.

### Note
Given that this could be considered a breaking change it would probably be wise to take a decision on this before a stable 1.0 is released.

### Environment
MacOS latest
Nuekit version & JS runtime: Nue 1.0.0-RC.4 • Bun 1.2.19

PR #600, opened 2025-08-01
feat: Atom feed generation

### LINKED ISSUE #283: RSS Feed Generation

NickSdot:
Closes: #283. This is a follow up to, and based on, https://github.com/nuejs/nue/pull/556.

**Differences:**
- added nested collections feed generation
- added missing fields like `subtitle` (prev. description), `author`
- added `has_feed` option. It works with inerhitance, childs can be (in/ex)cluded on demand.
- decoupled generation form `getPageData()`
- proper handling of inherited data like author
- handles dates properly
- added tests

**Discussion:**
- Because generation is opt-in via `has_feed`, I don't think adding a `--feed` flag is neccessary. When do you ever want to deploy without making sure your feed is updated? But your call, @tipiirai. Fwiw, performance lgtm.
- As reported in https://github.com/nuejs/nue/issues/598 collection links are curreny ambigious. In this implementation I decided for the variant with trailing slash. I think it is out of scope for this PR to find a better solution, and collection links should append `.html`
- As reported in https://github.com/nuejs/nue/issues/599 build directories are not fully wiped on new builds. This requires workaround like [this](https://github.com/nuejs/nue/compare/master...NickSdot:nuejs-nue:pr-556?expand=1#diff-6585d999f9c77ac7f564455cd5453affa6f93a990c0253f2660e6d66f9c0ba51R200) to ensure a feed is no longer in the build if e.g. `has_feed` was changed to `false`. The issue, however, applies to normal files as well, not only feeds.

--- 

@nobkd I forked your branch and kept your commits so that you don't lose co-authorship.


NickSdot:
I have a `<link rel="alternate" type="application/atom+xml"/>` implementation ready, too. Not 100% sure if it is how you with more core experience would do it. Wan't me to commit that here or prefer a separate PR later?

PR #614, opened 2025-09-20
Fix typo: remove repeated 'Modern React'

kappalasaimohith:
Removed a repeated phrase in the text
Before:
       Modern React Modern React represents a common approach...
After:
     Modern React represents a common approach...

Just wanted to check if that works for you. Let me know if you'd prefer it to be left as is!


nobkd:
thanks

PR #617, opened 2025-09-28
Update html-syntax.md

seneca:
These three lines run into one line when viewed in the browser. In theory a single carriage return should divide the line but it doesn't here. Two carried returns create a paragraph and it displays here correctly as separate lines.

### DIFF
--- a/packages/www/docs/html-syntax.md
+++ b/packages/www/docs/html-syntax.md
@@ -275,7 +275,9 @@ this.mount(name, target, data)
 ```
 
 **name** - Component name
+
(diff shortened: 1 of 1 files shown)


nobkd:
> In theory a single carriage return should divide the line

No, markdown ignores single line breaks outside of code blocks. (github changes this behavior in at least its issue / pr descriptions) You can get a break with two spaces at the end of the line (adds a `<br>` I think) , or with an empty line (new `<p>` block). 

---

I'm not sure, if in this case, it would be more sensible to just use list items? :shrug:

PR #626, opened 2025-10-14
Update layout-system.md

seneca:
I think the "main" section is missing from the list. See the above picture in the rendered .md file.


nobkd:
ty

PR #622, opened 2025-10-10
fix: win create

nobkd:
fix #620.
Windows does not have `unzip`, but can unzip with `tar`.
(at least on my linux distro, `tar` can't unzip, though)

PR #632, opened 2025-10-15
refactor: enhance unzip function for cross-platform support

marcogomez-dev:
fix #622 

Improve the unzip function in create.js to use `tar` on Windows and `unzip` on Linux/macOS, ensure target directory creation, add numbered steps with comments for clarity, and implement proper cleanup of temporary ZIP files. Update tests to handle nested directory extraction and sort file lists for consistent comparison. These changes enable reliable template unzipping across platforms without external dependencies.

### DIFF
--- a/packages/nuekit/src/cmd/create.js
+++ b/packages/nuekit/src/cmd/create.js
@@ -1,11 +1,10 @@
-
-import { mkdir } from 'node:fs/promises'
+import { mkdir, unlink } from 'node:fs/promises'
 import { join } from 'node:path'
+import os from 'node:os'
 
 export const NAMES = 'blog full minimal spa'.split(' ')
 
 export async function create(name, { dir, baseurl }) {
-
   if (!name) return console.log('❌ USAGE: nue create <template-name>')
 
   if (!NAMES.includes(name)) {
@@ -27,23 +26,24 @@ export async function create(name, { dir, baseurl }) {
     console.log(`   nue\n`)
 
     return true
-
   } catch (error) {
     console.error(`❌ ${error.message}`)
   }
 }
 
-
 export async function getLocalZip(name, dir) {
   const path = join(dir, `${name}.zip`)
-  if (!await Bun.file(path).exists()) throw new Error(`${path} not found`)
+  if (!(await Bun.file(path).exists())) throw new Error(`${path} not found`)
   console.log(`📦 Using local template: ${path}`)
   return Bun.file(path)
 }
 
 // download from github
 //
-export async function fetchZip(name, baseurl='https://github.com/nuejs/nue/raw/master/packages/templates') {
+export async function fetchZip(
+  name,
+  baseurl = 'https://github.com/nuejs/nue/raw/master/packages/templates'
+) {
   const url = `${baseurl}/${name}.zip`
   const resp = await fetch(url)
   if (resp.status != 200) throw new Error(`${url} not found`)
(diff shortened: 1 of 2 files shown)


nobkd:
Hello, could you provide an example command where my changes in #622 didn't fix the unzip issue / extracted wrongly? I just tried creating all templates (on windows ofc) from the pr branch, and it worked without any issue.

marcogomez-dev:
> Hello, could you provide an example command where my changes in #622 didn't fix the unzip issue / extracted wrongly? I just tried creating all templates (on windows ofc) from the pr branch, and it worked without any issue.

Hi nobkd,

My testing method focused on running the unit tests with `bun test`👀, not manual template creation 😵. The failure occurs specifically in the test within **`packages\nuekit\test\cmd\create.test.js`**.

### Diagnosis: Windows Unzip Behavior

My research into using `tar` and `unzip` on Windows suggests a key difference in how directory structures are handled:

  * **`tar`** creates the exact directory structure contained within the ZIP archive.
  * **`unzip`** (when used without the explicit **`-e`** flag) only extracts files if the destination directory names precisely match the directory names inside the zipped file.

This is what I suspect is happening 🤔 because the test verifies `minimal/[files]`, but my result with `tar` was `minimal/minimal/[files]`.

As a temporary measure, I added the **`-e`** flag to the `unzip` command, along with a `sort` function. This was a **workaround** to make both `tar` and `unzip` behave similarly, allowing me to resolve file comparison inconsistencies and continue working.

### Specific Error from `bun test`

I have re-tested your PR, and here is the exact error output from the failing `unzip` test:

```bash
cmd\create.test.js:
✓ getLocalZip
» fetchZip
22 | test('unzip', async () => {
23 |    const dir = 'minimal'
24 |    const zip = await getLocalZip(dir, 'cmd')
25 |    await unzip(dir, zip)
26 |    const files = await readdir(dir)
27 |    expect(files).toEqual([ "index.html", "index.css" ])
                       ^
error: expect(received).toEqual(expected)

  [
-   "index.html",   # This file is either missing or extracted incorrectly
    "index.css",
+   "index.html",   # Expected to be found here
  ]

- Expected  - 1
+ Received  + 1

      at <anonymous> (C:\Repos\nue\packages\nuekit\test\cmd\create.test.js:27:17)
✗ unzip [63.00ms]
```

After seeing better this specific error, it seems likely that applying a **sorting mechanism for the extracted files** will resolve the current test failure, and the `-e` flag may actually be unnecessary.

Let me know if you need any more information.

nobkd:
There's no `-e` flag on `unzip`.

<details>

```md
UNZIP(1)         General Commands Manual        UNZIP(1)

NAME
       unzip  -  list, test and extract compressed files
       in a ZIP archive

SYNOPSIS
       unzip   [-Z]   [-cflptTuvz[abjnoqsCDKLMUVWX$/:^]]
       file[.zip]    [file(s) ...]     [-x xfile(s) ...]
       [-d exdir]
```

</details>

---

Thanks for the notice on the tests, I checked that part again, and windows really just orders files differently, therefore I updated my PR, and fixed the test.

---

The directory to be created is already contained inside the zip, therefore we don't need to ensure the directory (goes for both unzip and win tar), and if we created extra directories, we would get results like `[...]/minimal/minimal/[...]`.

---

If it's alright with you, I would close this pr, as there was already #622 before which just needed a test correction.

marcogomez-dev:
> There's no `-e` flag on `unzip`.
> 
> ```
> UNZIP(1)         General Commands Manual        UNZIP(1)
> 
> NAME
>        unzip  -  list, test and extract compressed files
>        in a ZIP archive
> 
> SYNOPSIS
>        unzip   [-Z]   [-cflptTuvz[abjnoqsCDKLMUVWX$/:^]]
>        file[.zip]    [file(s) ...]     [-x xfile(s) ...]
>        [-d exdir]
> ```
> 
> Thanks for the notice on the tests, I checked that part again, and windows really just orders files differently, therefore I updated my PR, and fixed the test.
> 
> The directory to be created is already contained inside the zip, therefore we don't need to ensure the directory (goes for both unzip and win tar), and if we created extra directories, we would get results like `[...]/minimal/minimal/[...]`.
> 
> If it's alright with you, I would close this pr, as there was already #622 before which just needed a test correction.

Hi!! :D

Yep, I'm agree with all

Only for knowledge purpose, I take the `-e` flag from,
https://www.geeksforgeeks.org/linux-unix/unzip-command-in-linux/

nobkd:
> Only for knowledge purpose, I take the -e flag from,
> https://www.geeksforgeeks.org/linux-unix/unzip-command-in-linux/

Not sure where they have that from, I didn't find it in the `man` pages or in the `unzip` command's help :man_shrugging: 

---

Edit:

They even contradict themselves?

```
-e

Extracts files from the ZIP archive even while keeping the directory structure. This is useful if you want to keep the equal directory hierarchy as in the archive.

unzip -e [filename.zip]
```

vs

```
7. Encryption and Security

    To extract encrypted files and maintain security, appoint the '-e' alternative:

Syntax

unzip -e filename.zip

Example

    If 'archive.zip' incorporates encrypted files, use the '-e' alternative for stable extraction.

Syntax and Output

unzip -e archive.zip

    The '-e' option ensures the secure extraction of encrypted files from 'archive.zip,' maintaining the confidentiality and safety of the contents.
```

maybe AI hallucination? (because there's password protection `6. Password-Protected ZIP` breforehand?)

PR #636, opened 2025-10-18
Fix self-closing img tag in author section

seneca:
Incomplete img closing tag.

### DIFF
--- a/packages/www/docs/html-file-types.md
+++ b/packages/www/docs/html-file-types.md
@@ -45,7 +45,7 @@ Create reusable components and [layout modules](/docs/layout-system) for server-
 </footer>
 
 <author class="card">
-  <img src="{ avatar }" alt="{ name }">
+  <img src="{ avatar }" alt="{ name }" />
   <h3>{ name }</h3>
   <p>{ role }</p>
 


nobkd:
HTML5 doesn't care about that slash. There are no real self-closing tags, only void tags, and those just *allow* the "self-closing" slash.

I personally would prefer to just never use `/>` for void tags (just `<void>`), as the usage imitates XML and may cause confusion, as real self-closing behavior does not exist in html5 (different from xml/svg), but that's just my opinion. :shrug: 

Of course the docs should be consistent in the use of that pattern.

what do you think?

seneca:
Hi nobkd,

This is a tricky one.
The author emphasises that the framework wants to be as close to the standard as possible.
The closing tag /> is part of the HTML standard.
Should this be differed to the author to decide?
All in all this is a very minor thing anyway.

nobkd:
From the HTML standard:

> 6. Then, if the element is one of the [void elements](https://html.spec.whatwg.org/multipage/syntax.html#void-elements), or if the element is a [foreign element](https://html.spec.whatwg.org/multipage/syntax.html#foreign-elements), then there ***may be*** a single U+002F SOLIDUS character (/), which on [foreign elements](https://html.spec.whatwg.org/multipage/syntax.html#foreign-elements) marks the start tag as self-closing. ***On [void elements](https://html.spec.whatwg.org/multipage/syntax.html#void-elements)***, it does not mark the start tag as self-closing but instead ***is unnecessary and has no effect*** of any kind. For such void elements, it should be ***used only with caution*** — especially since, if directly preceded by an [unquoted attribute value](https://html.spec.whatwg.org/multipage/syntax.html#unquoted), it becomes part of the attribute value rather than being discarded by the parser.
> - <https://html.spec.whatwg.org/multipage/syntax.html#start-tags>, emphasis is mine

My opinion on just not using `/>` in HTML stays the same, but in the end it doesn't really matter, as it --- as you said --- is such a minor thing.
And yes, I think @tipiirai should decide on that.

PR #639, opened 2025-10-21
simpler version import, use builtin cli text styling

nobkd:


tipiirai:
This is awesome!  Didn't know about styleText and the package.json import is clever.

Issue #649, opened 2025-11-12
draggable  is enumerated and not Boolean and eventhandler will change details tag's open attribute to false


lyg1949:
### Describe the Bug
when set draggable to "true", nue will set only draggable (no ="true"),

eventhandler(like onclick) will change details tag's open attribute to false

### Environment
- OS: Linux
- Nuekit 2.0

code like below:

```
<!doctype dhtml>

<body>
  <main>
    <details>
      <summary>
          <span class="tree-item">dir1</span>
      </summary>
      <details>
          <summary draggable="true" :ondragstart="drag" :onclick="click">
              <span class="tree-item">file1</span>
          </summary>
      </details>
      
  </details>
  <details>
    <summary>
        <span class="tree-item">dir2</span>
    </summary>
    <details>
        <summary>
            <span class="tree-item">file2</span>
        </summary>
    </details>
    
</details>
  </main>

  <script>

async drag(e){
console.log("ondragstart")
}
async  click(e){
e.preventDefault();
console.log(e)
}
</script>

</body>
```

PR #658, opened 2026-03-31
fix bug

lyg1949:
1. dragable is not boolean
2. if attribute is bot bind, when ui fresh , it's value will be set to default( even if it is changed by user)


lyg1949:
#649

